PR #1, opened 2015-01-16
Birthday

agclever:
This is an initial version of a script that copies Mixpanel OAuth data to Redshift via S3. See https://clever.atlassian.net/browse/INFRA-571 on more context.

Tested:
Ran command and verified that it works.

```
$ MIXPANEL_API_KEY=<mixpanel_key> \
MIXPANEL_API_SECRET=<mixpanel_secret> \
AWS_ACCESS_KEY_ID=<aws_key> \ 
AWS_SECRET_ACCESS_KEY=<aws_secret> 
AWS_REGION='us-east-1' \
AWS_REDSHIFT_HOST=<redshift_host> \
AWS_REDSHIFT_PORT=5439 \
AWS_REDSHIFT_USER=<user> \
AWS_REDSHIFT_PASSWORD=<password> \
AWS_REDSHIFT_DB_NAME=dev \
godep go run mixpanel_to_redshift.go \
-jsonpathsfile=<jsonpaths_file> \
-mixpanel_events=<events_csv> \
-exportdir=<export_dir> \
-redshifttable=<redshift_table>
```


agclever:
cc: @azylman @clever/go for Go paradigms input.

mohit:
Do we really want to check in all the godeps?

agclever:
Raf argues at https://github.com/Clever/gearman-load-logger/pull/1#discussion_r22510041 that it should be. I separated out the commits for easier review, but if you feel strongly I can separate that into another branch.

mixpanel_oauth_to_redshift.go
```diff
@@ -0,0 +1,177 @@
...
+			log.Panic(err)
+		}
+		defer db.Close()
+		err = CopyJsonDataFromS3ToRedshift(
+			db, *redshift_table, export_file, aws_access_key_id, aws_secret_access_key, *jsonpaths_file, aws_region)
+		if err != nil {
+			log.Panic(err)
+		}
+	}
+}
```
rgarcia: Is there any way to make this script idempotent? If this script is not carefully scheduled (with consideration for the date on the machine it runs, etc.), then we risk copying duplicate data into redshift.

mixpanel_oauth_to_redshift.go
```diff
@@ -0,0 +1,177 @@
...
+	"io/ioutil"
+	"log"
+	"net/http"
+	"net/url"
+	"os"
+	"sort"
+	"time"
+
+	"github.com/Clever/pathio"
+	_ "github.com/lib/pq" // Postgres Driver
```
rgarcia: IMO obvious comment

mixpanel_oauth_to_redshift.go
```diff
@@ -0,0 +1,177 @@
...
+		time.Now().AddDate(0, 0, -1).Format("2006-01-02"),
+		"Date in YYYY-MM-DD format. Defaults to yesterday.")
+	mixpanel_export_dir = flag.String("mixpanel_export_dir", "s3://clever-mixpanel/OAuth",
+		"Directory to store the exported mixpanel data.")
+	mixpanel_export_request_timeout = flag.Duration(
+		"mixpanel_export_request_timeout", 10*time.Minute,
+		"Timeout value to be used while making a mixpanel export request. Defaults to 10 mins.")
+	redshift_table       = flag.String("redshift_table", "mixpanel_oauth", "Name of the redshift table.")
+	export_from_mixpanel = flag.Bool("export_from_mixpanel", true, "Whether to export from mixpanel.")
+	copy_to_redshift     = flag.Bool("copy_to_redshift", true, "Whether to copy to redshift.")
```
rgarcia: not sure if this convention extends to flags, but in general snake case is avoided in Go

Could shorten some of these, e.g. `copy`, `export`, `table`

mohit:
@agclever @rgarcia after looking at the conversation from `gearman-load-logger` I'm still not convinced that checking in godeps is the right way to go here. The biggest issues are bloated commits and making it harder to track actual changes in repo. However in the past I've also seen issues with caching of dependencies as an issue that comes up with checked-in godeps. Would it not be possible to just write our Makefile in a way so that the godeps are cleaned up/pulled down when running tests locally or in drone?

azylman:
Here's a reference to the caching issue Mohit is talking about: https://groups.google.com/forum/#!msg/golang-nuts/NtZtHx_dr0o/Hd9ygOhStvMJ

Basically, because of the way godeps constructs your GOPATH, if you have godeps X, someone changes them to Y, and you pull, all of your builds, runs, etc. will still be using X unless you delete your Go build caches.

mixpanel_oauth_to_redshift.go
```diff
@@ -0,0 +1,177 @@
...
+	aws_access_key_id     = os.Getenv("AWS_ACCESS_KEY_ID")
+	aws_secret_access_key = os.Getenv("AWS_SECRET_ACCESS_KEY")
+	aws_region            = os.Getenv("AWS_REGION")
+	mixpanel_api_key      = os.Getenv("MIXPANEL_API_KEY")
+	mixpanel_api_secret   = os.Getenv("MIXPANEL_API_SECRET")
+	redshift_host         = os.Getenv("AWS_REDSHIFT_HOST")
+	redshift_port         = os.Getenv("AWS_REDSHIFT_PORT")
+	redshift_database     = os.Getenv("AWS_REDSHIFT_DB_NAME")
+	redshift_user         = os.Getenv("AWS_REDSHIFT_USER")
+	redshift_password     = os.Getenv("AWS_REDSHIFT_PASSWORD")
```
nathanleiby: Recommended to exit immediately if env not set. You can use `Get` in https://github.com/segmentio/go-env/blob/master/env.go#L7 and handle error

mixpanel_oauth_to_redshift.go
```diff
@@ -0,0 +1,177 @@
...
+func CopyJsonDataFromS3ToRedshift(
+	db *sql.DB, redshift_table string, export_file string, aws_access_key_id string,
+	aws_secret_access_key string, jsonpaths_file string, aws_region string) error {
+	copy_cmd := fmt.Sprint(
+		"COPY ", redshift_table, " FROM '", export_file, "' WITH ",
+		"credentials ", fmt.Sprintf("'%s=%s;%s=%s'", "aws_access_key_id", aws_access_key_id, "aws_secret_access_key", aws_secret_access_key),
+		" json '", jsonpaths_file,
+		"' region '", aws_region,
+		"' timeformat 'epochsecs'")
+	log.Print("Executing command: ", copy_cmd)
```
nathanleiby: Don't log credentails - more info on why: https://github.com/Clever/dev-handbook/blob/master/logging.md#security

mixpanel_oauth_to_redshift.go
```diff
@@ -0,0 +1,177 @@
...
+	str_value, ok := value.(string)
+	if !ok {
+		json_value, err := json.Marshal(value)
+		if err != nil {
+			return "", err
+		}
+		str_value = string(json_value)
+	}
+	return str_value, nil
+}
```
nathanleiby: You've done a nice job breaking things into simple methods. Recommend adding unit tests for each (`StringOrJsonMarshal`, `ComputeSig`, `Request`)

mixpanel_oauth_to_redshift.go
```diff
@@ -0,0 +1,177 @@
...
+	hash := md5.Sum([]byte(joined_args + m.api_secret))
+	return hex.EncodeToString(hash[:]), nil
+}
+
+func (m *MixpanelExport) Request(method string, params map[string]interface{}) ([]byte, error) {
+	params["api_key"] = m.api_key
+	expire_time := time.Now().Add(*mixpanel_export_request_timeout)
+	params["expire"] = expire_time.Unix()
+	params["format"] = "json"
+	delete(params, "sig")
```
nathanleiby: Might deserve a comment on why you're deleting this. (would it just be user error to pass "sig" as a param?)

mixpanel_oauth_to_redshift.go
```diff
@@ -0,0 +1,177 @@
...
+		"' region '", aws_region,
+		"' timeformat 'epochsecs'")
+	log.Print("Executing command: ", copy_cmd)
+	_, err := db.Exec(copy_cmd)
+	return err
+}
+
+func main() {
+	flag.Parse()
+	export_file := fmt.Sprintf("%s/%s", *mixpanel_export_dir, *mixpanel_export_date)
```
nathanleiby: could use `filepath.Join` http://golang.org/pkg/path/filepath/

README.md
```diff
@@ -1 +1,22 @@
...
+Mixpanel into Redshift via S3.
+
+If the repository grows to include more use cases, this should be changed to a more of a
+plugin/mix-in model.
+
+## Running
+
+This script should be run daily using a cron job.
+
+TODO: details
```
nathanleiby: Make sure you detail

- necessary env
- several commands showing variety of flag usage

Also include information on how to build / test

rgarcia:
@mohit my response to bloated commits was in that discussion: "The only bloated commits are when you change dependencies (rare) and the initial switch to godep (happens once) so I don't see much of a cost there.." Can you respond to that?

Also, your suggestion of instrumenting the makefile to accommodate lack of Godeps in git is the kind of complexity I expressed wariness about in that thread: "...trying to vendor without really vendoring ends up complicating build processes and local development since you have a bunch of extra unknown
state (i.e. the content of your empty vendor directory)." Do you agree?

Re: "harder to track actual changes in repo", I disagree. For one, changing dependencies happens very, very infrequently. And when these changes occur, they should be in separate commits from logical (or "real") changes. I don't see how that makes it more difficult to track changes. In fact, I think it makes it easier to understand when code we haven't written (which is the majority of the code in a repo like this one) changes.

The cost of the case linked to in the google group seems real, albeit rare (again, we rarely change dependencies). Perhaps clearing the go build cache could happen in a `make clean` target? Also, this problem (managing the state of built code) seems like it would be present regardless of whether we committed dependencies or not. I would also consider managing the state of this cache an example of my point above about managing variable state across environments. At least Godep removes the state of the code from the equation.

mixpanel_oauth_to_redshift.go
agclever: Changed all the golang names to mixedCaps in 7c05f5ac2c721fc5c4ea8dab5b1b73d0ea795d7c. The go tools seems to use all lowercase, no separator convention for its flags so I changed the flag names accordingly in 83dd471.

mixpanel_oauth_to_redshift.go
agclever: I would argue that is needed. A package is being imported and not being used so there is no way of knowing what it does. Plus the import path does not provide any information.

mixpanel_oauth_to_redshift.go
agclever: Done in 9cec7a347c9e51a3f0353b86cc994e7ce9cdf8b2.

mixpanel_oauth_to_redshift.go
agclever: Good catch. Done in 9d9d4323c57fb4bba4d8a30fec8716ed5b10be76

mixpanel_oauth_to_redshift.go
agclever: While I agree on the utility of unit tests. This repo is in a state of flux right now so I would prefer adding tests when it is in a more stable state.

mixpanel_oauth_to_redshift.go
agclever: I tried that but strips out one of the `/` from the s3 path breaking the filename.

README.md
agclever: Again will prefer adding this when I have a better idea of what the repository would look like.

mohit:
@rgarcia 

re: bloated commits vs makefile complexity: This is where my experience has been different. We do change/update dependencies quite often and each time that happens additional changes in the PR make code review more prone to errors (missing things, mental fatigue). If others are okay with this cost as opposed to the added complexity in the Makefile then it seems good to me. Personally I'd like to try working without checking in the godeps and only move to that when we actually hit any issues with that behavior. 

In essence I've seen dependencies changing often and felt the pain of trying to review with lots of non-clever code, but not really felt the pain of local development issue due to not checked in godeps. That makes me feel like we might be adding code without actually needing to. If there are actual occurrences of issues that we've seen with  godeps when they are not checked in then I'm just speaking from a lack of experience with it.

mixpanel_oauth_to_redshift.go
agclever: Done in 37cfe2f. It is not exactly user error, but the signature needs to be recomputed since the params have changed.

rgarcia:
@mohit moving forward can you flag the problem of reviewing being difficult due to including deps in git? I'd like to see it in the wild since I personally have not felt the pain outside of the cases I've mentioned here (birthday PRs and PRs that switch to Godep). It could also be a symptom of people not making separate PRs for dependency upgrades.

If you find it agreeable I'd like to steer this discussion away from the bikeshed by

1. Continuing to collect data about costs this may incur. E.g. flagging problems with the approach when they happen.

2. Sticking to the [recommended](https://github.com/golang/go/wiki/PackageManagementTools#vendoring) way of vendoring, i.e. committing the code.

mixpanel_oauth_to_redshift.go
```diff
@@ -0,0 +1,177 @@
...
+
+	for k := range params {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+
+	joined_args := ""
+	for _, k := range keys {
+		joined_args += k
+		value, err := StringOrJsonMarshal(params[k])
```
rgarcia: I've seen [a different](https://github.com/austinchau/go-mixpanel/blob/4e0c0d6f76663facb6414ed55b17bea1cc444715/mixpanel.go) approach in a mixpanel library I've used before that does not have a need for a `StringOrJsonMarshal`. At first glance it seems to stem from requiring the user to pass `params` as a `map[string]string` instead of `map[string]interface{}`. Might be worth considering to simplify this code a bit.

Godeps/Godeps.json
```diff
@@ -0,0 +1,27 @@
+{
+	"ImportPath": "github.com/Clever/redshifter",
+	"GoVersion": "go1.3.3",
```
rgarcia: why not version 1.4?

Godeps/_workspace/src/github.com/Clever/pathio/Makefile
```diff
@@ -0,0 +1,29 @@
+SHELL := /bin/bash
+PKG = github.com/Clever/pathio
+PKGS := $(PKG)
+READMES = $(addsuffix /README.md, $(PKGS))
```
rgarcia: `PKG` and `READMES` can be `:=`

Godeps/_workspace/src/github.com/Clever/pathio/Makefile
```diff
@@ -0,0 +1,29 @@
...
+
+.PHONY: test golint README
```
rgarcia: `$(PKGS)` is also phony

`README` is not a target

`golint` can be made un-phony if you wish. See the approach in the stoked Makefile

rgarcia: `docs` is phony

mohit:
:+1:

mixpanel_oauth_to_redshift.go
agclever: I prefer this approach. Requiring the user to pass params as a `map[string]string` means the user has to take care of the marshaling which would result in duplicated effort.

Godeps/Godeps.json
agclever: My machine did not have go1.4 so godep defaulted to using 1.3.3. Fixed in dafa1dfc283fdc7137f5ef0ae0b456725f4bb5a6.

Godeps/_workspace/src/github.com/Clever/pathio/Makefile
agclever: This is the makefile of clever/pathio which is an dependency. I assume that I shouldn't make any changes here.

Godeps/_workspace/src/github.com/Clever/pathio/Makefile
agclever: same as above. This is an import.

README.md
```diff
@@ -1 +1,22 @@
...
+
+## Motivation
+
+[AWS Redshift](http://aws.amazon.com/redshift/) is a columnar storage based data warehouse solution.
+It is optimized for analysis and business intelligence and has integration with many tools.
+
+## Features
+
+Right now this repository contains a single script that exports OAuth authorize events data from
+Mixpanel into Redshift via S3.
```
nathanleiby: Nothing sensitive here but do take care in public repo to minimize references to Clever specific info.

mixpanel_oauth_to_redshift.go
```diff
@@ -0,0 +1,179 @@
...
+
+func main() {
+	flag.Parse()
+	exportFile := fmt.Sprintf("%s/%s", *mixpanelExportDir, *mixpanelExportDate)
+
+	if *exportFromMixpanel {
+		mixpanelExport := NewMixpanelExport()
+		log.Println("Exporting mixpanel data for", *mixpanelExportDate)
+		params := map[string]interface{}{
+			"event":     [1]string{"authorize"},
```
nathanleiby: Is this the only thing that's schema specific? Might we parametrize this so we can easily pull other mixpanel events, instead of hard-coding?

mixpanel_oauth_to_redshift.go
```diff
@@ -0,0 +1,179 @@
...
+	awsSecretAccessKey = env.MustGet("AWS_SECRET_ACCESS_KEY")
+	awsRegion          = env.MustGet("AWS_REGION")
+	mixpanelAPIKey     = env.MustGet("MIXPANEL_API_KEY")
+	mixpanelAPISecret  = env.MustGet("MIXPANEL_API_SECRET")
+	redshiftHost       = env.MustGet("AWS_REDSHIFT_HOST")
+	redshiftPort       = env.MustGet("AWS_REDSHIFT_PORT")
+	redshiftDatabase   = env.MustGet("AWS_REDSHIFT_DB_NAME")
+	redshiftUser       = env.MustGet("AWS_REDSHIFT_USER")
+	redshiftPassword   = env.MustGet("AWS_REDSHIFT_PASSWORD")
+	jsonpathsFile      = flag.String("jsonpathsfile", "s3://clever-mixpanel/OAuth.json",
```
nathanleiby: again, wouldn't include this clever-specific path/usage in a public repo (`s3://clever-mixpanel/OAuth`)

agclever: Done in e2dd5c2e3dd5d2a17136ef62a56ef399181f40da.

mixpanel_oauth_to_redshift.go
agclever: Done in eb75d41.

Godeps/_workspace/src/github.com/Clever/pathio/Makefile
rgarcia: derp :blush:

README.md
rgarcia: :+1: seems easy to remove references to oauth here and in the filename `mixpanel_oauth_to_redshift.go`

rgarcia:
also @mohit @azylman good prezo on dependencies I saw on golang-nuts a couple of days ago: https://www.youtube.com/watch?v=yi5A3cK1LNA#t=1058. Linked to the final point made by the presenter, which is basically that seeing external dependencies in your vendor directory makes you think twice about bringing in external dependencies and pushes you to write simpler programs. I've personally fallen into this trap, and seen us as a team falling into this trap when pulling in npm dependencies without giving them much thought, only to find out later that they were nowhere near production-ready. Vendoring is one way to enforce thinking about dependencies before bringing them in.

agclever:
@rgarcia Couldn't find your original comment about removing more references, but this is now done in 832565af38d22f184e5505eb33803182a6560ccf and 064bd8ab9c0704a515f09d02b8f363ab6682f25c.

mixpanel_oauth_to_redshift.go
agclever: We need some way of determining whether the database has the data for a particular day or not. One way to do that is  to store a state in some file whenever a copy command for a particular date finishes. These improvements, along with others can come in a follow up PR, possibly along with the change to get analytics DB into redshift.

agclever:
Offline LGTM from @nathanleiby. Merging.

PR #2, opened 2015-01-23
Script to copy data from postgres to redshift.

agclever:
The motivation for this change is at https://clever.atlassian.net/browse/INFRA-598.

Tested: Ran the two scripts to make sure they perform the intended operations.


agclever:
WIP. To do before merging:
- Unit tests (possibly integration too).
- Flag validation.

agclever:
@nathanleiby Unit tests have been added for all modules. Please take a look.

nathanleiby:
@agclever thanks for heads up! Will take a look asap

nathanleiby:
Go specific:

- run go lint to see some go patterns patterns you could do (e.g. method docstrings as missing for public methods)

Clever conventions:

- improve readme - most importantly, document usage - more info in [dev handbook](https://github.com/Clever/dev-handbook/blob/master/documentation.md#what-goes-in-a-readme)
- add a .drone.yml file to automate tests
- add a VERSION file and use gitsem to version
- consider adding a Makefile to specify commands for things like testing, linting, and generating docs. One exmaple: [Optimus makefile](https://github.com/Clever/optimus/blob/master/Makefile)
- make it a standalone deployable (@mohit can support or cover this; model: https://github.com/Clever/gearman-load-logger)
  - Dockerfile
  - launch folder

mixpanel/mixpanel_export_test.go
```diff
@@ -0,0 +1,75 @@
...
+		"api_key": "TESTMIXPANELAPIKEY",
+		"expire":  12312324324,
+		"events":  []string{"event1", "event2"},
+		"param1":  "value1",
+	}
+	m, exp := MixpanelExport{apiSecret: "TESTMIXPANELAPISECRET"}, "1beb8d4f61da24302844e252a8ff6e75"
+	sig, err := m.ComputeSig(params)
+	if err != nil {
+		t.Errorf("Unexpected error %s for ComputeSign(%+v). Expected %s", err.Error(), params, exp)
+	}
```
nathanleiby: A lot of our code uses github.com/stretchr/testify ... one liners like `assert.NoError` are handy

mixpanel/mixpanel_export_test.go
```diff
@@ -0,0 +1,75 @@
...
+	if sig != exp {
+		t.Errorf("Unexpected value %s for ComputeSign(%+v). Expected %s", sig, params, exp)
+	}
```
nathanleiby: Could be `assert.Equal`.

One reason I push for this library is readable output, including printing of the expected and actual values

postgres_to_redshift.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/segmentio/go-env"
+)
+
+var (
+	// TODO: include flag validation
+	awsRegion  = env.MustGet("AWS_REGION")
+	s3prefix   = flag.String("s3prefix", "", "s3 path to be used as a prefix for temporary storage of postgres data")
+	tables_csv = flag.String("tables", "", "Tables to copy as CSV")
+	dumppg     = flag.Bool("dumppostgres", true, "Whether to dump postgres")
+	updateRS   = flag.Bool("updateredshift", true, "Whether to replace redshift")
```
nathanleiby: Thoughts on env vs flags? It seems inconsistent. AWS key, secret, and region are from ENV. All else is from flags.

I think the long run discussion is here for getting to consistent env/flag patterns: https://clever.atlassian.net/browse/INFRA-503

One reason I mention it, too, is that I'm thinking of `postgres/postgres.go` as a library, so to me the NewDB constructor would make more sense to take arguments instead of read from flags. It's very convenient here, but is specific to being used in as a CLI app.

mohit: there is a library being worked on at https://github.com/Clever/go-utils/pull/1 that might be helpful

mohit:
I would skip `standalone-izing` from this PR. Something that can be added later.

postgres_to_redshift.go
agclever: @nathanleiby I prefer flags wherever possible because the go support for them is much better. The only reason AWS key, secret and region are from ENV is because these are used by the pathio library so it makes sense to re-use the same here.

agclever: @mohit I know about the go-utils library. In fact, I wrote that PR motivated by the discussion here.

mixpanel/mixpanel_export_test.go
agclever: Done in 026b438f1b81ce38c49e2213ac87fc5c6484b670, 871b6619ac04fbbafb1a1f4a0cdeb2c9a9bad30d, and 41604af8cccd3771ad9fadf20cb35a0eab72c1a8.

mixpanel/mixpanel_export_test.go
agclever: Done in 026b438, 871b661, and 41604af.

postgres_to_redshift.go
agclever: As for the using flags/arguments, eventually everything would have to run using the commandline. The only thing we would lose is the ability to talk to multiple instances of postgres/redshift.

agclever:
* Removed the postgres_to_redshift script in 4993e2186c6f9dbbe83616d0d7e812fe9cb61181. This should be moved to a separate repo.
* Fixed golint errors in a730c420c4a9ca1c136a1527bd084c3c5131e910.
* Added makefile in eb9fcf5.
* Added .drone.yml in 128c3b3.
* Updated readme in b3c1bba and 3dc8530.
* Used gitsem in c538b1e.

.drone.yml
```diff
@@ -0,0 +1,15 @@
+image: go1.4
+script:
+  - mkdir -p $GOPATH/src/github.com/Clever
+  - ln -s `pwd` $GOPATH/src/githuv.com/Clever/redshifter
```
nathanleiby: s/githuv/github

agclever: Fixed in 6d30cb6.

nathanleiby:
lgtm

rgarcia:
question for people on this PR: should we use godocdown for public repos? Compare https://github.com/Clever/redshifter/blob/master/mixpanel/README.md with http://godoc.org/github.com/Clever/redshifter/mixpanel.

It seems to me like the only value of godocdown is for private repos that we couldn't use godoc.org with. (Although even with private repos you can clone them into your $GOPATH and run `godoc -http=":8080"` or `godoc github.com/Clever/<name of repo>`).

cc @azylman for thoughts (I think you originally introduced godocdown?)

azylman:
I like documentation to be in as many places as possible, as long as it's
kept up to date (which godocdown enforces) but I don't care too much either
way.

I don't know for sure, but I think we could have our own godoc server
inside vpc that has access to our private repos.
If we don't like godocdown, we should run our own godoc server, but I don't
see it a question of private vs public as much as a question of
documentation in multiple places for convenience or one place so there's a
canonical location.

PR #3, opened 2015-02-19
redshift: drop table if exists

nathanleiby:
If the destination table hasn't yet been created in Redshift, you
don't need to drop it. This failure to drop causes and error
and causes the github.com/clever/postgres-to-redshift to fail
on initial import (data gets imported but only to 'temp' table
name)

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -3,6 +3,8 @@ script:
   - go get github.com/tools/godep
   - mkdir -p $GOPATH/src/github.com/Clever
   - ln -s `pwd` $GOPATH/src/github.com/Clever/redshifter
+  - export AWS_ACCESS_KEY_ID=x
+  - export AWS_SECRET_ACCESS_KEY=x
   - make test
 notify:
   email:
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,5 +1,6 @@
 image: bradrydzewski/go:1.3
 script:
+  - go get github.com/tools/godep
   - mkdir -p $GOPATH/src/github.com/Clever
   - ln -s `pwd` $GOPATH/src/github.com/Clever/redshifter
   - make test
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,4 +1,4 @@
-image: go1.4
+image: bradrydzewski/go:1.3
 script:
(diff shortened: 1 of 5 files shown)


nathanleiby:
@agclever I just enabled CI in drone, but it's not yet working fully -- likely need to tweak `.drone.yml`

agclever:
LGTM once drone passes.

nathanleiby:
@agclever would you mind taking a look at this test failure in mixpanel request output? I have seen this locally too:

https://drone.ops.clever.com/github.com/clever/redshifter/commit/f9caf64f7e4c06fde6ce2d8dcbe562d5034af94a?branch=redshift-drop-table-if-exists

agclever:
The test passes locally for me. The only difference I can think of is that we are using go1.4 locally go1.3 on drone. What version do you have installed?

```
(redshift-drop-table-if-exists) $ make
.....
.....
=== RUN TestRequest
2015/02/19 22:07:53 https://data.mixpanel.com/api/2.0/method/?api_key=apikey&events=%5B%22event1%22%2C%22event2%22%5D&expire=10000600&format=json&param1=value1&param2=value2&sig=825125f9640ce6bee7b6c25fd33eacb3
--- PASS: TestRequest (0.00s)
PASS
ok  	github.com/Clever/redshifter/mixpanel	0.015s
```

nathanleiby:
Seems likely it is a go1.3. issue, since that's what I'm running locally. I'll get updated to go1.4 and check on that. 

Maybe they've changed some aspects of reflection that affect `assert.Equal`? In the [go1.4 release notes](http://tip.golang.org/doc/go1.4) I see: "The reflect package's Type interface has a new method, Comparable, that reports whether the type implements general comparisons." which may be the cause. 

The following change gets things passing in go1.3 
https://github.com/Clever/redshifter/commit/fe27d42617162650004e338c950361499687a693 .. though I then realized also needed to pass some env  https://github.com/Clever/redshifter/commit/2adbab81a62c5cd9cabc0468d757bbc758ffe92b

agclever:
That is probably the case. I am in favor of incorporating fe27d42 in this PR since we don't have a go1.4 image for drone yet.

agclever:
LGTM.

PR #5, opened 2015-02-26
Fix the ALTER TABLE command when refreshing redshift tables

agclever:
Rename does not allow setting the schema in the new name.

Tested: Ran the actual command.


nathanleiby:
:+1:

PR #6, opened 2015-02-27
Grant select and references priviliges to public after refresh

agclever:
Without the grant query, users no longer have access once a table gets refreshed

Tested: Unit tests and ran the query locally.


agclever:
Merging this now because it is a small change. Feel free to comment though.

nathanleiby:
:+1:

PR #7, opened 2015-03-18
Refresh redshift without deleting tables

agclever:
Previously the refresh tables operation deleted the old tables and replaced them with new ones. This means that we can't do optimizations like foreign keys and perform a certain set of operations (recreating views, granting access) after every sync.

Instead, this change will use tmp tables to replace all the data within the tables during the refresh while still keeping the actual table objects and the settings intact.

Tested: Unit tests and ran postgres-to-redshift worker for a single table to make sure that it actually performs the operations correctly.


redshift/redshift.go
```diff
@@ -73,40 +73,34 @@ func (r *Redshift) CopyJSONDataFromS3(schema, table, file, jsonpathsFile, awsReg
...
-	}
-	if c.PrimaryKey {
-		attributes = append(attributes, "SORTKEY")
-		constraints = append(constraints, "PRIMARY KEY")
-	}
-	if c.NotNull {
-		constraints = append(constraints, "NOT NULL")
-	}
-	return fmt.Sprintf("%s %s %s %s", c.Name, c.ColType, strings.Join(attributes, " "), strings.Join(constraints, " "))
+func (r *Redshift) createTempTable(tmpname, schema, name string) error {
```
mohit: Should we add a comment here about why a temp table is being used?

redshift/redshift.go
```diff
@@ -114,20 +108,16 @@ func (r *Redshift) createTable(schema, name string, ts postgres.TableSchema) err
 // and later renaming the temporary table to the original one.
```
mohit: fix comment.

mohit:
LGTM barring the comments

mohit:
Also we should ensure that the copy from temp table to real table works for large tales

agclever:
@mohit updated comments and changed the implementation to use actual tables in a different schema. See c9fce23.

redshift/redshift.go
```diff
@@ -73,78 +73,74 @@ func (r *Redshift) CopyJSONDataFromS3(schema, table, file, jsonpathsFile, awsReg
...
 			}
 			group.Done()
 		}(name, ts)
 	}
+	if err := group.Wait(); err != nil {
+		group.Error(err)
+	}
+	if _, err := r.logAndExec(fmt.Sprintf(`DROP SCHEMA "%s" CASCADE`, *tmpschema), false); err != nil {
+		group.Error(err)
+	}
```
mohit: should we have a cleanup in case of error?

mohit:
LGTM

PR #8, opened 2015-03-24
Increase PoolSize for postgres DB to 10 from the default 5

agclever:
For some reason, the driver is not great at releasing the connections or waiting for connections to be released if already full. This causes the program to hang if we dump more than 5 tables in parallel. Increasing the connection pool size to fix it.

Tested: Ran the program locally with the increased pool size and verified that it works.

Driver code relating poolsize of 5: https://github.com/go-pg/pg/blob/v2.1/db.go#L66


nathanleiby:
10 likewise seems arbitrary - might run into this problem later.

What if instead you allowed passing in a config struct to `NewDB`, and in postgres-to-redshifter it sets `PoolSize` equal to `len(tables)`?

https://github.com/Clever/postgres-to-redshift/blob/master/main.go#L30

nathanleiby:
Alternately could investigate further why Pool connections are not becoming available? That might be a problem with driver itself or could be how we're calling it. (maybe the s3writers aren't getting closed or are hanging onto the connection)

agclever:
@nathanleiby Added a struct in 8d5ab83

agclever:
Also I tried to see if it was something about us calling it. The driver code is supposed to close the writer and even us calling it did not make a difference. `Close` is a no-op for our case so it should not make a difference. It wasn't apparent why the lock condition variable wasn't being released so I didn't spend too much time diving into it.

nathanleiby:
lgtm

PR #13, opened 2015-08-04
allow multiple simultaneous redshifter jobs from providing different temp schemas

schimmy:
also update the godeps to be consistent with other repos- Clever should be capitalized


schimmy:
I'm not sure how the flags here would get parsed, but in any case I think this is safe. Other PR forthcoming for `redshift-to-postgres`, the only consumer of that exported method

agclever:
LGTM.

rzendacott:
I think I got assigned here by accident - reassigning to Schimmy.

PR #14, opened 2015-08-19
Vacuum analyze each individual table after refresh

agclever:
Running a full vacuum on the entire redshift database is an very expensive query. This changes redshifter to run a vacuum and analyze on a specific table as the last step of a refresh.

Tested: Unit tests.


bstein-clever:
Is it really called "Vacuum full" if it's just a single table? *sigh*
LGTM otherwise.

agclever:
The `FULL` implies that the vacuum should sort and delete. http://docs.aws.amazon.com/redshift/latest/dg/r_VACUUM_command.html

PR #16, opened 2015-09-25
Modify redshift for use by s3 to redshift, part 1

schimmy:
This is the first of a few PRs to integrate my `s3->redshift` work into this repo. I'm next going to add in a bunch of `getSQL` type commands and refactor the refresh table function, but I wanted to get this out in stages so it was easier to review.

### DIFF
--- a/cmd/mixpanel_to_redshift.go
+++ b/cmd/mixpanel_to_redshift.go
@@ -24,14 +24,14 @@ var (
 		"Date in YYYY-MM-DD format. Defaults to yesterday.")
 	mixpanelExportDir = flag.String("exportdir", "", "Directory to store the exported mixpanel data.")
 	host              = flag.String("host", "", "Address of the redshift host")
-	port              = flag.Int("port", 0, "Address of the redshift host")
+	port              = flag.String("port", "5439", "Port of the redshift host")
 	db                = flag.String("database", "", "Redshift database to connect to")
 	user              = flag.String("user", "", "Redshift user to connect as")
 	schema            = flag.String("schema", "public", "Schema with the redshift table.")
 	table             = flag.String("table", "", "Name of the redshift table.")
 	pwd               = flag.String("password", "", "Password for the redshift user")
-	timeout           = flag.Duration("connecttimeout", 10*time.Second,
-		"Timeout while connecting to Redshift. Defaults to 10 seconds.")
+	redshiftTimeout   = flag.Int("connecttimeout", 10,
+		"Timeout in seconds while connecting to Redshift. Defaults to 10 seconds.")
 	exportFromMixpanel = flag.Bool("export", true, "Whether to export from mixpanel.")
 	copyToRedshift     = flag.Bool("copy", true, "Whether to copy to redshift.")
 )
@@ -58,7 +58,7 @@ func main() {
 	}
 
 	if *copyToRedshift {
-		r, err := redshift.NewRedshift(*host, *port, *db, *user, *pwd, int(*redshiftTimeout.Seconds()))
+		r, err := redshift.NewRedshift(*host, *port, *db, *user, *pwd, *redshiftTimeout)
 		defer r.Close()
 		if err != nil {
 			log.Fatal(err)
--- a/cmd/mixpanel_to_redshift.go
+++ b/cmd/mixpanel_to_redshift.go
@@ -16,7 +16,6 @@ import (
 
 var (
 	// TODO: include flag validation
-	awsRegion          = env.MustGet("AWS_REGION")
 	jsonpathsFile      = flag.String("jsonpathsfile", "", "s3 file with jsonpaths data.")
 	mixpanelEvents     = flag.String("mixpanelevents", "", "Comma separated values of events to be exported.")
 	mixpanelExportDate = flag.String("exportdate",
@@ -58,12 +57,17 @@ func main() {
 	}
 
 	if *copyToRedshift {
-		r, err := redshift.NewRedshift(*host, *port, *db, *user, *pwd, *redshiftTimeout)
+		s3Info := redshift.S3Info{
+			Region:    env.MustGet("AWS_REGION"),
+			AccessID:  env.MustGet("AWS_ACCESS_KEY_ID"),
+			SecretKey: env.MustGet("AWS_SECRET_ACCESS_KEY"),
+		}
+		r, err := redshift.NewRedshift(*host, *port, *db, *user, *pwd, *redshiftTimeout, s3Info)
 		defer r.Close()
 		if err != nil {
 			log.Fatal(err)
 		}
-		if err := r.CopyJSONDataFromS3(*schema, *table, exportFile, *jsonpathsFile, awsRegion); err != nil {
+		if err := r.CopyJSONDataFromS3(*schema, *table, exportFile, *jsonpathsFile); err != nil {
 			log.Fatal(err)
 		}
 		if err := r.VacuumAnalyze(); err != nil {
--- a/redshift/README.md
+++ b/redshift/README.md
@@ -18,15 +18,16 @@ redshift database.
 #### func  NewRedshift
 
 ```go
-func NewRedshift(host, port, db, user, pwd string, timeout int) (*Redshift, error)
+func NewRedshift(host, port, db, user, pwd string, timeout int, s3Info S3Info) (*Redshift, error)
 ```
 NewRedshift returns a pointer to a new redshift object using configuration
-values passed in on instantiation and the AWS env vars we assume exist
+values passed in on instantiation and the AWS env vars we assume exist Don't
+need to pass s3 info unless doing a COPY operation
 
 #### func (*Redshift) CopyGzipCsvDataFromS3
 
 ```go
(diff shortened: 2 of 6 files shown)


schimmy:
@agclever I'm actually going to assign to @bstein-clever so that we can move more quickly on this, but feel free to weigh in as well.

cmd/mixpanel_to_redshift.go
```diff
@@ -16,15 +16,21 @@ import (
...
-	mixpanelExportDir  = flag.String("exportdir", "", "Directory to store the exported mixpanel data.")
-	redshiftSchema     = flag.String("redshiftschema", "public", "Schema with the redshift table.")
-	redshiftTable      = flag.String("redshifttable", "", "Name of the redshift table.")
+	mixpanelExportDir = flag.String("exportdir", "", "Directory to store the exported mixpanel data.")
+	host              = flag.String("host", "", "Address of the redshift host")
+	port              = flag.String("port", "5439", "Port of the redshift host")
+	db                = flag.String("database", "", "Redshift database to connect to")
+	user              = flag.String("user", "", "Redshift user to connect as")
+	schema            = flag.String("schema", "public", "Schema with the redshift table.")
+	table             = flag.String("table", "", "Name of the redshift table.")
```
bstein-clever: Nit: password, here (I'd also accept `pass`). I keep reading this as something to do with the working directory. :(

redshift/README.md
```diff
@@ -18,15 +18,16 @@ redshift database.
 #### func  NewRedshift
 
 ```go
-func NewRedshift() (*Redshift, error)
+func NewRedshift(host, port, db, user, pwd string, timeout int, s3Info S3Info) (*Redshift, error)
```
bstein-clever: Ditto on password.

redshift/redshift.go
```diff
@@ -19,59 +17,55 @@ type dbExecCloser interface {
...
+
 // Redshift wraps a dbExecCloser and can be used to perform operations on a redshift database.
 type Redshift struct {
 	dbExecCloser
-	accessID  string
-	secretKey string
+	s3Info S3Info
 }
 
 var (
```
bstein-clever: Do we need this var block at all if all the contents are diappearing?

redshift/redshift.go
```diff
@@ -19,59 +17,55 @@ type dbExecCloser interface {
...
-	return &Redshift{sqldb, awsAccessKeyID, awsSecretAccessKey}, nil
+	return &Redshift{sqldb, s3Info}, nil
 }
 
-func (r *Redshift) logAndExec(cmd string, creds bool) (sql.Result, error) {
+func (r *Redshift) logAndExec(cmd string) (sql.Result, error) {
 	log.Print("Executing Redshift command: ", cmd)
-	if creds {
-		cmd += fmt.Sprintf(" CREDENTIALS '%s=%s;%s=%s'", "aws_access_key_id", r.accessID, "aws_secret_access_key", r.secretKey)
-	}
```
bstein-clever: We had out logs printing out the credentials? :(

redshift/redshift.go
```diff
@@ -79,19 +73,22 @@ func (r *Redshift) CopyGzipCsvDataFromS3(schema, table, file, awsRegion string,
...
 	copyCmd := fmt.Sprintf(
 		`COPY "%s"."%s" (%s) FROM '%s' WITH REGION '%s' GZIP CSV DELIMITER '%c'`,
-		schema, table, strings.Join(cols, ", "), file, awsRegion, delimiter)
-	copyCmd += " IGNOREHEADER 0 ACCEPTINVCHARS TRUNCATECOLUMNS TRIMBLANKS BLANKSASNULL EMPTYASNULL DATEFORMAT 'auto' ACCEPTANYDATE COMPUPDATE ON"
-	_, err := r.logAndExec(copyCmd, true)
+		schema, table, strings.Join(cols, ", "), file, r.s3Info.Region, delimiter)
+	opts := " IGNOREHEADER 0 ACCEPTINVCHARS TRUNCATECOLUMNS TRIMBLANKS BLANKSASNULL EMPTYASNULL DATEFORMAT 'auto' ACCEPTANYDATE COMPUPDATE ON"
+	creds := fmt.Sprintf(` CREDENTIALS 'aws_access_key_id=%s;aws_secret_access_key=%s'`, r.s3Info.AccessID, r.s3Info.SecretKey)
+	fullCopyCmd := fmt.Sprintf("%s%s%s", copyCmd, opts, creds)
+	_, err := r.logAndExec(fullCopyCmd)
```
bstein-clever: Nit: can we make this the same style as the CopyJSONDataFromS3 function?
Either
```
opts = <bar>
creds = <foo>
cmd = <qux>
%s%s%s cmd,opts,creds
```

or
```
creds = <foo>
copy <opts> %s creds
```

Rather than mixing both? It's a bit more confusing to read about.

bstein-clever:
Some small nits, but otherwise this looks good.
Thanks for the cleanup!

redshift/redshift.go
schimmy: hmm, suppose not, although I'm about to add more to it in the next pr! so I guess thank you for the catch, but in this specific case it's all right :)

bstein-clever: Ha, ok, though normally I might say it's a better practice to remove it entirely then add it back in. That confuses something like git-blame less!

redshift/redshift.go
agclever: The logs never printed out credentials. The logging happens before the conditional. This affects the actual command that needs to be run.

redshift/redshift.go
```diff
@@ -19,59 +17,55 @@ type dbExecCloser interface {
...
+func (r *Redshift) CopyJSONDataFromS3(schema, table, file, jsonpathsFile string) error {
+	creds := fmt.Sprintf(`CREDENTIALS 'aws_access_key_id=%s;aws_secret_access_key=%s'`, r.s3Info.AccessID, r.s3Info.SecretKey)
 	copyCmd := fmt.Sprintf(
-		`COPY "%s"."%s" FROM '%s' WITH json '%s' region '%s' timeformat 'epochsecs' COMPUPDATE ON`,
-		schema, table, file, jsonpathsFile, awsRegion)
-	_, err := r.logAndExec(copyCmd, true)
+		`COPY "%s"."%s" FROM '%s' WITH json '%s' region '%s' timeformat 'epochsecs' COMPUPDATE ON %s`,
+		schema, table, file, jsonpathsFile, r.s3Info.Region, creds,
+	)
+	_, err := r.logAndExec(copyCmd)
```
agclever: This will cause credentials to be logged.

redshift/redshift.go
```diff
@@ -19,59 +17,55 @@ type dbExecCloser interface {
 	Exec(query string, args ...interface{}) (sql.Result, error)
 }
 
+// S3Info holds the information necessary to copy data from s3 buckets
+type S3Info struct {
+	Region    string
```
agclever: The region is a property of the s3 bucket being read from/copied to and not the redshift cluster. In the sense that we need to set `aws_region` params if the s3 bucket being used is in a different region than the redshift cluster. The value for this variable might be different based on the bucket you are interfacing with.

redshift/redshift.go
```diff
@@ -79,19 +73,22 @@ func (r *Redshift) CopyGzipCsvDataFromS3(schema, table, file, awsRegion string,
 	}
 	copyCmd := fmt.Sprintf(
 		`COPY "%s"."%s" (%s) FROM '%s' WITH REGION '%s' GZIP CSV DELIMITER '%c'`,
-		schema, table, strings.Join(cols, ", "), file, awsRegion, delimiter)
-	copyCmd += " IGNOREHEADER 0 ACCEPTINVCHARS TRUNCATECOLUMNS TRIMBLANKS BLANKSASNULL EMPTYASNULL DATEFORMAT 'auto' ACCEPTANYDATE COMPUPDATE ON"
-	_, err := r.logAndExec(copyCmd, true)
+		schema, table, strings.Join(cols, ", "), file, r.s3Info.Region, delimiter)
+	opts := " IGNOREHEADER 0 ACCEPTINVCHARS TRUNCATECOLUMNS TRIMBLANKS BLANKSASNULL EMPTYASNULL DATEFORMAT 'auto' ACCEPTANYDATE COMPUPDATE ON"
```
agclever: Should we add `STATUPDATE` to this set of options?

redshift/redshift.go
bstein-clever: Ahh, I'd read this as log.Sprintf, not fmt.Sprintf. That makes much more sense.

redshift/redshift.go
schimmy: hmm, this is a huge pain, I might take this time to go back to calling the copy functions directly, but start using go's actual transaction library instead of calling begin and end transaction as well, which solves some of the pain that returning the sql was solving...

redshift/redshift.go
schimmy: good point - this will actually be pretty easy to fix when I do the final 3/3 PR here, but good feedback!

schimmy:
@ag vastly changed all this in PR #18, closing this one.

PR #17, opened 2015-09-28
Stage more modifications before merging

schimmy:
We now will return SQL instead of just executing the statements.

Tests are a little weird, but I kept the old ones around. not sure the best way to improve those since they basically are testing if we concatenate strings together correctly, with a tiny bit of logic along with them.

Finally, I'm not sure that the mixpanel loader will work in its current form, so I'm thinking about removing it. Unfortunately we can't be sure whether anyone uses it or not, however. Abhinav suggests that anyone using it can stay vendored to the old version, and we would welcome PRs to add it back if someone is using it and can spend the time to make sure it works.

### DIFF
--- a/redshift/README.md
+++ b/redshift/README.md
@@ -18,7 +18,7 @@ redshift database.
 #### func  NewRedshift
 
 ```go
-func NewRedshift(host, port, db, user, pwd string, timeout int, s3Info S3Info) (*Redshift, error)
+func NewRedshift(host, port, db, user, password string, timeout int, s3Info S3Info) (*Redshift, error)
 ```
 NewRedshift returns a pointer to a new redshift object using configuration
 values passed in on instantiation and the AWS env vars we assume exist Don't
--- a/redshift/README.md
+++ b/redshift/README.md
@@ -34,7 +34,9 @@ type Meta struct {
 
 Meta holds information that might be not in Redshift or annoying to access in
 this case, we want to know the schema a table is part of and the column which
-corresponds to the timestamp at which the data was gathered
+corresponds to the timestamp at which the data was gathered NOTE: this will be
+useful for the s3-to-redshift worker, but is currently not very useful same with
+the yaml info
 
 #### type Redshift
 
@@ -55,29 +57,6 @@ NewRedshift returns a pointer to a new redshift object using configuration
 values passed in on instantiation and the AWS env vars we assume exist Don't
 need to pass s3 info unless doing a COPY operation
 
-#### func (*Redshift) GetCSVCopySQL
-
-```go
-func (r *Redshift) GetCSVCopySQL(schema, table, file string, ts Table, delimiter rune, creds, gzip bool) string
-```
-GetCSVCopySQL copies gzipped CSV data from an S3 file into a redshift table.
-
-#### func (*Redshift) GetJSONCopySQL
-
-```go
-func (r *Redshift) GetJSONCopySQL(schema, table, filename, jsonPaths string, creds, gzip bool) string
-```
-GetJSONCopySQL copies JSON data present in an S3 file into a redshift table. if
-not using jsonPaths, set to "auto"
-
-#### func (*Redshift) GetTruncateSQL
-
-```go
-func (r *Redshift) GetTruncateSQL(schema, table string) string
-```
-GetTruncateSQL simply returns SQL that deletes all items from a table, given a
-schema string and a table name
-
 #### func (*Redshift) RefreshTables
 
 ```go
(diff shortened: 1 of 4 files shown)


bstein-clever:
Re: mixpanel: I thought the recent emails from Raf suggested that we were trying to remove all uses of it?

redshift/README.md
```diff
@@ -18,36 +49,53 @@ redshift database.
...
 
-#### func (*Redshift) CopyJSONDataFromS3
+#### func (*Redshift) GetTruncateSQL
 
 ```go
-func (r *Redshift) CopyJSONDataFromS3(schema, table, file, jsonpathsFile string) error
+func (r *Redshift) GetTruncateSQL(schema, table string) string
 ```
-CopyJSONDataFromS3 copies JSON data present in an S3 file into a redshift table.
+GetTruncateSQL simply returns SQL that deletes all items from a table, given a
```
bstein-clever: Nit: we don't have simply in any of the other descriptions.

redshift/README.md
```diff
@@ -18,36 +49,53 @@ redshift database.
...
 ```
 RefreshTables refreshes multiple tables in parallel and returns an error if any
 of the copies fail.
 
+#### func (*Redshift) SafeExec
+
+```go
+func (r *Redshift) SafeExec(sqlIn []string) error
+```
+SafeExec allows execution of SQL in a transaction block While it seems a little
```
bstein-clever: Nit: "transaction block.\n While"

redshift/README.md
```diff
@@ -77,3 +125,42 @@ type S3Info struct {
...
 
 S3Info holds the information necessary to copy data from s3 buckets
+
+#### type SortableColumns
+
+```go
+type SortableColumns []ColInfo
+```
+
+Just a helper to make sure the CSV copy works properly
```
bstein-clever: Nit "A helper ..."

bstein-clever: Whoops. hit enter too soon. If it's just a helper, do we need to export it?

redshift/redshift.go
```diff
@@ -30,17 +30,55 @@ type Redshift struct {
...
+	Schema         string `yaml:"schema"`
+}
+
+// ColInfo is a struct that contains information about a column in a Redshift database.
+// SortKey and DistKey only make sense for Redshift
+type ColInfo struct {
+	Ordinal    int    `yaml:"ordinal"`
+	Name       string `yaml:"dest"`
+	Type       string `yaml:"type"`
+	DefaultVal string `yaml:"defaultval"`
```
bstein-clever: Are we intentionally mixing snake case (data_date_column) above with ... nothing case here (defaultval)?

redshift/redshift.go
```diff
@@ -30,17 +30,55 @@ type Redshift struct {
...
+	NotNull    bool   `yaml:"notnull"`
+	PrimaryKey bool   `yaml:"primarykey"`
+	DistKey    bool   `yaml:"distkey"`
+	SortOrd    int    `yaml:"sortord"`
+}
+
+// Just a helper to make sure the CSV copy works properly
```
bstein-clever: Nit: Just

redshift/redshift.go
```diff
@@ -49,68 +87,76 @@ func NewRedshift(host, port, db, user, pwd string, timeout int, s3Info S3Info) (
 }
 
 func (r *Redshift) logAndExec(cmd string) (sql.Result, error) {
-	log.Print("Executing Redshift command: ", cmd)
+	log.Print("Executing Redshift command: ", cmd) // TODO: filter out creds, perhaps
```
bstein-clever: Agreed to this todo, though I'd have no idea how to remove creds from the cmd, unless we make "logAndExec" take in a command and creds separately. Actually, that's not a bad idea.

redshift/redshift.go
```diff
@@ -49,68 +87,76 @@ func NewRedshift(host, port, db, user, pwd string, timeout int, s3Info S3Info) (
...
+	if gzip {
+		gzipSQL = "GZIP"
+	}
 	copyCmd := fmt.Sprintf(
-		`COPY "%s"."%s" FROM '%s' WITH json '%s' region '%s' timeformat 'epochsecs' COMPUPDATE ON %s`,
-		schema, table, file, jsonpathsFile, r.s3Info.Region, creds,
-	)
-	_, err := r.logAndExec(copyCmd)
-	return err
+		`COPY "%s"."%s" FROM '%s' WITH %s JSON '%s' REGION '%s' TIMEFORMAT 'auto' COMPUPDATE ON %s`,
```
bstein-clever: I really like this style of returning the SQL command. Nice idea!

redshift/redshift.go
```diff
@@ -49,68 +87,76 @@ func NewRedshift(host, port, db, user, pwd string, timeout int, s3Info S3Info) (
...
-		schema, table, strings.Join(cols, ", "), file, r.s3Info.Region, delimiter)
-	opts := " IGNOREHEADER 0 ACCEPTINVCHARS TRUNCATECOLUMNS TRIMBLANKS BLANKSASNULL EMPTYASNULL DATEFORMAT 'auto' ACCEPTANYDATE COMPUPDATE ON"
-	creds := fmt.Sprintf(` CREDENTIALS 'aws_access_key_id=%s;aws_secret_access_key=%s'`, r.s3Info.AccessID, r.s3Info.SecretKey)
-	fullCopyCmd := fmt.Sprintf("%s%s%s", copyCmd, opts, creds)
-	_, err := r.logAndExec(fullCopyCmd)
-	return err
+		`COPY "%s"."%s" (%s) FROM '%s' WITH REGION '%s' %s CSV DELIMITER '%c'`,
+		schema, table, strings.Join(colStrings, ", "), file, r.s3Info.Region, gzipSQL, delimiter)
+	opts := "IGNOREHEADER 0 ACCEPTINVCHARS TRUNCATECOLUMNS TRIMBLANKS BLANKSASNULL EMPTYASNULL DATEFORMAT 'auto' ACCEPTANYDATE COMPUPDATE ON"
+	fullCopyCmd := fmt.Sprintf("%s %s %s", copyCmd, opts, credSQL)
```
bstein-clever: Again, I'd love to see this match the style of GetJSONCopySQL, or vice versa, rather than having one split out the cmd, opts, and creds, and pass them differently.

Also, if creds are just pasted into the end always, if we keep it separate, we don't even have to pass them into this routine, and then can more easily supply them as a separate argument to logAndExec.

redshift/redshift.go
```diff
@@ -49,68 +87,76 @@ func NewRedshift(host, port, db, user, pwd string, timeout int, s3Info S3Info) (
...
-	}
-	_, err := r.logAndExec(strings.Join(cmds, "; "))
-	return err
+// GetTruncateSQL simply returns SQL that deletes all items from a table, given a schema string and a table name
+func (r *Redshift) GetTruncateSQL(schema, table string) string {
+	return fmt.Sprintf(`DELETE FROM "%s"."%s"`, schema, table)
 }
 
-// RefreshTable refreshes a single table by first copying gzipped CSV data into a temporary table
+// RefreshTable refreshes a single table by copying gzipped CSV data into a temporary table
```
bstein-clever: Were't we no longer using temporary tables, and no longer using gzip?

bstein-clever:
Some nits and minor changes I'd like to make, bu otherwise this is shaping up nicely!

schimmy:
Closing in favor of #18

PR #18, opened 2015-09-30
Modify redshifter before including s3-to-redshift work

schimmy:
Closing #16 and #17 in favor of this one.

Big changes, mostly moving to golang sql library transactions instead of our poor-man's transactions :)

The only weird thing here is the `s3Info` struct, don't worry about that one @agclever and @bstein-clever , the PR when I modify for `s3-to-redshift` will have a different way of handling it that will probably make you happy.

### DIFF
--- a/cmd/mixpanel_to_redshift.go
+++ b/cmd/mixpanel_to_redshift.go
@@ -67,9 +67,14 @@ func main() {
 		if err != nil {
 			log.Fatal(err)
 		}
-		if err := r.CopyJSONDataFromS3(*schema, *table, exportFile, *jsonpathsFile); err != nil {
+		// do use s3 creds, don't use GZIP
+		// I am not confident this currently works - you may have to set timeformat to 'epochsecs'
+		copySQL := r.GetJSONCopySQL(*schema, *table, exportFile, *jsonpathsFile, true, false)
+		err = r.SafeExec([]string{copySQL})
+		if err != nil {
 			log.Fatal(err)
 		}
+
 		if err := r.VacuumAnalyze(); err != nil {
 			log.Fatal(err)
 		}
--- a/cmd/mixpanel_to_redshift.go
+++ b/cmd/mixpanel_to_redshift.go
@@ -69,9 +69,15 @@ func main() {
 		}
 		// do use s3 creds, don't use GZIP
 		// I am not confident this currently works - you may have to set timeformat to 'epochsecs'
-		copySQL := r.GetJSONCopySQL(*schema, *table, exportFile, *jsonpathsFile, true, false)
-		err = r.SafeExec([]string{copySQL})
+		tx, err := r.Begin()
 		if err != nil {
+			log.Fatal(err)
+		}
+		if err := r.RunJSONCopy(tx, *schema, *table, exportFile, *jsonpathsFile, true, false); err != nil {
+			tx.Rollback()
+			log.Fatal(err)
+		}
+		if err := tx.Commit(); err != nil {
 			log.Fatal(err)
 		}
 
--- a/cmd/mixpanel_to_redshift.go
+++ b/cmd/mixpanel_to_redshift.go
@@ -24,14 +24,14 @@ var (
 		"Date in YYYY-MM-DD format. Defaults to yesterday.")
 	mixpanelExportDir = flag.String("exportdir", "", "Directory to store the exported mixpanel data.")
 	host              = flag.String("host", "", "Address of the redshift host")
-	port              = flag.Int("port", 0, "Address of the redshift host")
+	port              = flag.String("port", "5439", "Port of the redshift host")
 	db                = flag.String("database", "", "Redshift database to connect to")
 	user              = flag.String("user", "", "Redshift user to connect as")
 	schema            = flag.String("schema", "public", "Schema with the redshift table.")
 	table             = flag.String("table", "", "Name of the redshift table.")
 	pwd               = flag.String("password", "", "Password for the redshift user")
-	timeout           = flag.Duration("connecttimeout", 10*time.Second,
-		"Timeout while connecting to Redshift. Defaults to 10 seconds.")
+	redshiftTimeout   = flag.Int("connecttimeout", 10,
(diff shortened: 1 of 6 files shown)


schimmy:
also while using a transaction for just one copy command would be overkill, I figure it's all right to force clients to create a transaction to allow us to have the power we need for `s3-to-redshift`

redshift/redshift.go
```diff
@@ -7,113 +7,180 @@ import (
...
+}
+
+// Meta holds information that might be not in Redshift or annoying to access
+// in this case, we want to know the schema a table is part of
+// and the column which corresponds to the timestamp at which the data was gathered
+// NOTE: this will be useful for the s3-to-redshift worker, but is currently not very useful
+// same with the yaml info
+type Meta struct {
+	DataDateColumn string `yaml:"datadatecolumn"`
+	Schema         string `yaml:"schema"`
```
agclever: Should `schema` just be a part of the `Table` struct?

redshift/redshift.go
```diff
@@ -7,113 +7,180 @@ import (
...
+// SortKey and DistKey only make sense for Redshift
+type ColInfo struct {
+	Ordinal    int    `yaml:"ordinal"`
+	Name       string `yaml:"dest"`
+	Type       string `yaml:"type"`
+	DefaultVal string `yaml:"defaultval"`
+	NotNull    bool   `yaml:"notnull"`
+	PrimaryKey bool   `yaml:"primarykey"`
+	DistKey    bool   `yaml:"distkey"`
+	SortOrd    int    `yaml:"sortord"`
```
agclever: Not sure what this means.

redshift/redshift.go
```diff
@@ -7,113 +7,180 @@ import (
...
+	if creds {
+		credSQL = `CREDENTIALS 'aws_access_key_id=?;aws_secret_access_key=?'`
+		credArgs = []interface{}{r.s3Info.AccessID, r.s3Info.SecretKey}
+	}
+	gzipSQL := ""
+	if gzip {
+		gzipSQL = "GZIP"
+	}
+	copySQL := `COPY "?"."?" FROM '?' WITH ? JSON '?' REGION '?' TIMEFORMAT 'auto' STATUPDATE COMPUPDATE ON %s`
+	copyStmt, err := r.Prepare(fmt.Sprintf(copySQL, credSQL))
```
agclever: Could we move the prepared queries outside the function? From what I understand, that's one of the major advantages of using prepared queries because we can re-use the same ones with different arguments. Also saves us from doing an extra hop on every invocation but I think that's somewhat minor advantage for this case.

redshift/redshift.go
```diff
@@ -7,113 +7,180 @@ import (
...
+	var credArgs []interface{}
+	if creds {
+		credSQL = `CREDENTIALS 'aws_access_key_id=?;aws_secret_access_key=?'`
+		credArgs = []interface{}{r.s3Info.AccessID, r.s3Info.SecretKey}
+	}
+	gzipSQL := ""
+	if gzip {
+		gzipSQL = "GZIP"
+	}
+	copySQL := `COPY "?"."?" FROM '?' WITH ? JSON '?' REGION '?' TIMEFORMAT 'auto' STATUPDATE COMPUPDATE ON %s`
```
agclever: I think this needs to be `STATUPDATE ON COMPUPDATE ON`
http://docs.aws.amazon.com/redshift/latest/dg/copy-parameters-data-load.html#copy-statupdate

redshift/redshift.go
```diff
@@ -7,113 +7,180 @@ import (
...
+	cols := sortableColumns{}
+	cols = append(cols, ts.Columns...)
+	sort.Sort(cols)
+	colStrings := []string{}
+	for _, ci := range cols {
+		colStrings = append(colStrings, ci.Name)
+	}
+
+	copySQL := fmt.Sprintf(`COPY "?"."?" (?) FROM '?' WITH REGION '?' ? CSV DELIMITER '?'`)
+	opts := "IGNOREHEADER 0 ACCEPTINVCHARS TRUNCATECOLUMNS TRIMBLANKS BLANKSASNULL EMPTYASNULL DATEFORMAT 'auto' ACCEPTANYDATE STATUPDATE COMPUPDATE ON"
```
agclever: Ditto. Should be `STATUPDATE ON`

redshift/redshift.go
```diff
@@ -7,113 +7,180 @@ import (
...
+	fullCopySQL := fmt.Sprintf("%s %s %s", copySQL, opts, credSQL)
+	copyStmt, err := r.Prepare(fullCopySQL)
```
agclever: Same comment re: prepare.

cmd/mixpanel_to_redshift.go
```diff
@@ -51,14 +57,30 @@ func main() {
...
+		}
+		if err := r.RunJSONCopy(tx, *schema, *table, exportFile, *jsonpathsFile, true, false); err != nil {
+			tx.Rollback()
 			log.Fatal(err)
 		}
+		if err := tx.Commit(); err != nil {
+			log.Fatal(err)
+		}
+
 		if err := r.VacuumAnalyze(); err != nil {
```
agclever: Shouldn't need this since we are now using `STATUPDATE ON` and `VACUUM`s are very disruptive.

redshift/redshift.go
```diff
@@ -7,113 +7,180 @@ import (
...
+	if err != nil {
+		return err
 	}
-	_, err := r.logAndExec(strings.Join(cmds, "; "), false)
+	_, err = tx.Stmt(truncStmt).Exec(schema, table)
 	return err
 }
 
-// RefreshTable refreshes a single table by first copying gzipped CSV data into a temporary table
+// RefreshTable refreshes a single table by copying gzipped CSV data into a temporary table
```
agclever: We should change this docstring to reflect the current behavior.

redshift/redshift.go
schimmy: hmm, my reading of this is that we don't actually get much benefit from it. However, using prepared statements here makes it a lot easier to test and safer to input data with, so it's still a benefit. Perhaps future work to improve it?

PR #19, opened 2015-10-06
s3-to-redshift worker

schimmy:
Actual PR for the s3-to-redshift worker. It removes the old `mixpanel` and `postgres` export functionality as we won't need that in the future and I'm pretty confident no one else uses this in the wild. This PR creates a worker which can pull from `s3` into `redshift` using `COPY` looking for the most recent file or a set date.

After reading a configuration file, the worker ensures the target `redshift` table exists and has the proper columns. If not, it performs the proper SQL commands to get into the state dictated by the configuration file. Then it attempts to find the most recent file on s3 matching a pattern. If the `s3` data is more recent than what is in `redshift`, it loads that `s3` data into `redshift`.


schimmy:
@agclever @nikhilpandit @drhurd

schimmy:
@drhurd ready for you, I can work with you if you have any questions as this is a huge change.

Additionally, I'm going to be working on two parts while you're reviewing the logic:
1) Tests - right now the tests don't work, I'm going to fix them
2) Making this a worker - with a dockerfile, etc

However I'd like to get started looking at the code together now :)

schimmy:
closing this in favor of #20

PR #20, opened 2015-10-12
Convert to s3 to redshift worker

schimmy:
Actual PR for the s3-to-redshift worker. It removes the old `mixpanel` and `postgres` export functionality as we won't need that in the future and I'm pretty confident no one else uses this in the wild. This PR creates a worker which can pull from s3 into redshift using `COPY` looking for the most recent file or a set date.

After reading a configuration file, the worker ensures the target redshift table exists and has the proper columns. If not, it performs the proper SQL commands to get into the state dictated by the configuration file. Then it attempts to find the most recent file on s3 matching a pattern. If the s3 data is more recent than what is in redshift, it loads that s3 data into redshift.


schimmy:
had to remove the "remove godeps" commit for space issues, but will add that back in before merge

cmd/s3_to_redshift.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	return nil
+}
+
+// This worker finds the latest file in s3 and uploads it to redshift
+// If the destination table does not exist, the worker creates it
+// If the destination table lacks columns, the worker creates those as well
+// The worker also uses a column in the data to figure out whether the s3 data is
+// newer than what already exists.
+func main() {
```
drhurd: Nitpicky, but should main be moved before the lower level functions in this file?

cmd/s3_to_redshift.go
```diff
@@ -0,0 +1,161 @@
...
+
+	timeout := 10 // can parameterize later if this is an issue
+	if host == "" {
+		host = "localhost"
+	}
+	if port == "" {
+		port = "5439"
+	}
+	db, err := redshift.NewRedshift(host, port, dbName, user, pwd, timeout)
+	fatalIfErr(err, "error getting redshift instance")
```
drhurd: Minor style issue: do we want to break out the s3 and redshift connection setup into their own functions? Purely for readability.

redshift/redshift.go
```diff
@@ -71,149 +67,296 @@ func (c sortableColumns) Less(i, j int) bool { return c[i].Ordinal < c[j].Ordina
...
 	}
-	copySQL := `COPY "?"."?" FROM '?' WITH ? JSON '?' REGION '?' TIMEFORMAT 'auto' STATUPDATE ON COMPUPDATE ON %s`
-	copyStmt, err := r.Prepare(fmt.Sprintf(copySQL, credSQL))
+
+	return fmt.Sprintf(" %s %s %s %s %s %s %s", c.Name, typeMapping[c.Type], defaultVal, notNull, sortKey, primaryKey, distKey)
+}
+
+// RunCreateTable runs the full create table command in the provided transaction, given a
+// redshift representation of the table.
+func (r *Redshift) RunCreateTable(tx *sql.Tx, table Table) error {
```
drhurd: Purely a style opinion, but why is the name `RunCreateTable` instead of just `CreateTable`? IMHO, this is usually used when you have a fully prepared operation, like the result of `tx.Prepare` in this case.

redshift/redshift.go
```diff
@@ -71,149 +67,296 @@ func (c sortableColumns) Less(i, j int) bool { return c[i].Ordinal < c[j].Ordina
...
+// RunUpdateTable figures out what columns we need to add to the target table based on the
+// input table, and completes this action in the transaction provided
+// Note: only supports adding columns currently, not updating existing columns or removing them
+func (r *Redshift) RunUpdateTable(tx *sql.Tx, targetTable, inputTable Table) error {
+	columnOps := []string{}
+	for _, inCol := range inputTable.Columns {
+		var existingCol ColInfo
+		// yes, n^2, but how many are we really going to have, and I want to keep the ordering
+		// ignore if there are extra columns
+		for _, targetCol := range targetTable.Columns {
```
drhurd: Alternative: construct a map of column_name -> *column. O(n) and maintains the order of inputTable.Columns. Might be unnecessary though.

redshift/redshift.go
```diff
@@ -71,149 +67,296 @@ func (c sortableColumns) Less(i, j int) bool { return c[i].Ordinal < c[j].Ordina
...
+	}
+
+	// TODO: use parameter placeholder syntax instead
+	existQueryFormat = `SELECT table_name
+			  FROM information_schema.tables WHERE table_schema='%s' AND table_name='%s'`
+
+	// returns one row per column with the attributes:
+	// name, type, default_val, not_null, primary_key, dist_key, and sort_ordinal
+	// need to pass a schema and table name as the parameters
+	schemaQueryFormat = `SELECT
```
drhurd: Should these be consts?

redshift/redshift.go
```diff
@@ -71,149 +67,296 @@ func (c sortableColumns) Less(i, j int) bool { return c[i].Ordinal < c[j].Ordina
...
-	opts := "IGNOREHEADER 0 ACCEPTINVCHARS TRUNCATECOLUMNS TRIMBLANKS BLANKSASNULL EMPTYASNULL DATEFORMAT 'auto' ACCEPTANYDATE STATUPDATE ON COMPUPDATE ON"
-	fullCopySQL := fmt.Sprintf("%s %s %s", copySQL, opts, credSQL)
-	copyStmt, err := r.Prepare(fullCopySQL)
-	if err != nil {
-		return err
-	}
-	args := []interface{}{schema, table, strings.Join(colStrings, ", "), file, r.s3Info.Region, gzipSQL, delimiter}
-	log.Printf("Running command: %s with args: %v", fullCopySQL, args)
-	_, err = tx.Stmt(copyStmt).Exec(append(args, credArgs...)...)
+	copySQL := fmt.Sprintf(`COPY "%s"."%s" FROM '%s' WITH %s JSON '%s' REGION '%s' TIMEFORMAT 'auto' STATUPDATE ON COMPUPDATE ON %s`, f.Schema, f.Table, f.GetDataFilename(), gzipSQL, f.JSONPaths, f.Bucket.Region(), credSQL)
```
drhurd: Should this string literal be a const? (similar to `existQueryFormat`)

redshift/redshift.go
```diff
@@ -71,149 +67,296 @@ func (c sortableColumns) Less(i, j int) bool { return c[i].Ordinal < c[j].Ordina
...
+	// can't use prepare b/c of redshift-specific syntax that postgres does not like
+	_, err := tx.Exec(fullCopySQL)
 	return err
 }
 
 // RunTruncate deletes all items from a table, given a transaction, a schema string and a table name
 // you shuold run vacuum and analyze soon after doing this for performance reasons
 func (r *Redshift) RunTruncate(tx *sql.Tx, schema, table string) error {
-	truncStmt, err := r.Prepare(`DELETE FROM "?"."?"`)
+	truncStmt, err := tx.Prepare(`DELETE FROM "?"."?"`)
```
drhurd: This one shouldn't need to be a const, but worth considering for consistent styling

redshift/redshift_test.go
```diff
@@ -41,84 +197,215 @@ func TestGetJSONCopySQL(t *testing.T) {
 	}
 }
 
-func TestCopyGzipCsvDataFromS3(t *testing.T) {
-	s3Info := S3Info{
-		Region:    "testregion",
-		AccessID:  "accesskey",
-		SecretKey: "secretkey",
+// test create and getColumnSQL at the same time
+// getColumnSQL is too simple to be worth testing independently
```
drhurd: Bold words.

s3filepath/s3filepath.go
```diff
@@ -0,0 +1,137 @@
...
+// We need to sort by data timestamp to find the most recent s3 file
+type byTimeStampDesc []s3.Key
+
+func (k byTimeStampDesc) Len() int           { return len(k) }
+func (k byTimeStampDesc) Swap(i, j int)      { k[i], k[j] = k[j], k[i] }
+func (k byTimeStampDesc) Less(i, j int) bool { return k[i].Key > k[j].Key } // reversing so we get sorted by date desc
+
+// FindLatestInputData looks for the most recent file matching the prefix
+// created by <schema>_<table>, using the RFC3999 date in the filename
+func FindLatestInputData(bucket Bucketer, schema, table, suppliedConf string, targetDate *time.Time) (*S3File, error) {
```
drhurd: While I don't have a better suggestion for a name, I think `FindLatestInputData` may be misleading. Splitting this into 2 separate functions, `FindLatestInputData`, and `FindInputDataForDate`, with an internal helper function encapsulating most of the logic, might be more clear.

cmd/s3_to_redshift.go
```diff
@@ -0,0 +1,161 @@
...
+
+		// figure out what the current state of the table is to determine if the table is already up to date
+		targetTable, lastTargetData, err := db.GetTableMetadata(inputConf.Schema, inputConf.Table, inputTable.Meta.DataDateColumn)
+		if err != nil && err != sql.ErrNoRows { // ErrNoRows is fine, just means the table doesn't exist
+			fatalIfErr(err, "Error getting existing latest table metadata") // use fatalIfErr to stay the same
+		}
+
+		// unless --force, don't update unless input data is new
+		if !inputConf.DataDate.After(*lastTargetData) {
+			if *force != true {
```
drhurd: *force == false

schimmy: that is not not a great idea

cmd/s3_to_redshift.go
schimmy: hmm, decided to pull out the table code instead

schimmy: actually this ended up being just as gross, I think I'm going to leave it here. It seems just a little too gross to have multiple functions, all of which need this run configuration

schimmy:
@drhurd the const question raf answered in slack, basically we always do vars but by convention it it's declared in the const block we don't modify them. The main problem is that you can't const a map, so then you have two sections and it gets awkward.

I'll take a look at the column map lookup, but your input on GetLatestInputData was VERY sharp, I was doing two separate things in the same function

cmd/s3_to_redshift.go
schimmy: it is below!

drhurd:
Sorry, my comment may have been miscommunicated. I was referring to the SQL queryFormat strings, not the map. If we are not planning on changing these, they seem like they should definitely be consts - relying on convention seems unnecessary when we can let the compiler enforce this rule.

drhurd:
Other than that, LGTM!

cmd/s3_to_redshift.go
drhurd: :+1: sounds good

PR #23, opened 2015-10-20
allow non-gzipped files as well

schimmy:
I need to be able to handle non-gzipped files as well for what @cgclever is working on, it was a TODO that I never got around to, although I actually prefer the smaller PR format :)


cgclever:
LGTM, looking forward to seeing some tangible data transfer

PR #27, opened 2015-10-26
int -> integer in type mapping

schimmy:
this was dumb


nikhilpandit:
LGTM!

schimmy:
basically, I tested this with create table but not with update table - I only tested that with `text`. Since `create table` accepts `int` as a synonym for `integer`, creation worked just fine. However, when looking at the metadata on the next run, we see `integer` as that's the non-aliased syntax.

This PR standardizes on the non-aliased syntax

PR #28, opened 2015-10-27
match existing functionality by truncating too-long columns

schimmy:
this is causing my inserts to fail :(


bstein-clever:
LGTM

PR #33, opened 2015-11-06
Fix copy from create table

schimmy:
Needed to make sure we didn't pick up on the config file if the config name happened to match the prefix of the data.

for instance:
`mongo_sis_config_x.yml`
`mongo_students_x.yml`
was working fine, but:
`mongo_apprequests_config_x.yml`
`mongo_apprequests_x.yml`
was confusing s3-to-redshift.


bstein-clever:
LGTM!

PR #34, opened 2015-11-13
IP-372 Remove ordinal check

schimmy:
This was me being a little overzealous. Since we don't use CSV copying
anymore, just JSON, ordering the columns is no longer needed.
Ensuring column order is just too strict.


carolinemodic:
lgtm

PR #39, opened 2015-12-14
Linting

natebrennand:
- adds `go vet` to Makefile
  - fixes lint errors
- separates test & build in drone config
- removes godoc generation
  - deletes pkg README's
  - adds godoc.org badge


Makefile
```diff
@@ -18,21 +20,18 @@ GODEP := $(GOPATH)/bin/godep
...
-ifneq ($(NOLINT),1)
-	@echo "LINTING..."
-	@$(GOLINT) $(GOPATH)/src/$@*/**.go
-	@echo ""
-endif
-	@echo "TESTING..."
-	@go test $@ -test.v
-	@echo ""
+	@echo "FORMATTING"
+	@gofmt -l=true $(GOPATH)/src/$@*/*.go
```
schimmy: wait, we don't want to overwrite? I assume `go vet` will fail and warn in that case?

schimmy:
LGTM other than question about `-l` instead of `-w` gofmt option

Makefile
natebrennand: Neither `golint` nor `go vet` warn if you're failing `gofmt`, it's a basic assumption that code is already compliant.
You should have your editor set up to format on save, let me know if you need help configuring it.

natebrennand: The `-l` flag lists all files in violation, `FGT` will make that return an error so that the build fails.
Previously we could have code on master that didn't follow `gofmt`.

PR #40, opened 2015-12-14
Fix date issue

schimmy:
This will just actually error when we hit the limit - we should figure out a better way to do this, but for now I've deleted some old files to get us under the limit.

I think we should store longer copies of the data for later processing in some smart way - right now we can only keep around about a month's worth of data before querying hits the 1000 item limit, and in any case that gets awkward over 1000 generally

Linked in the code, docs about 1000: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGET.html


nikhilpandit:
LGTM (after Slack discussion)

PR #42, opened 2016-01-11
increase gearcmd timeout for data copy

schimmy:
ideally this will take less time, but until we figure out how to address the issue, let's increase the timeout to 2 hours


bstein-clever:
Can we add a comment in the shell code for the next reader, and maybe a ticket so we don't forget to turn it back down?
LGTM otherwise.

PR #44, opened 2016-01-11
(Fixed the revert) Removes functionality finding latest file and uses…

alsmola:
This change un-reverts https://github.com/Clever/s3-to-redshift/pull/41 with some fixes to `pathio` (https://github.com/Clever/pathio/pull/13) and to the repo itself, including updating the README. Otherwise, it should be the same.


schimmy:
LGTM!

PR #45, opened 2016-01-13
Revert "increase gearcmd timeout for data copy"

bstein-clever:
Reverts Clever/s3-to-redshift#42

This didn't help, and I'd prefer the shorter timeout.
We're opening a ticket with Amazon about why this is failing, instead.


schimmy:
LGTM!

PR #47, opened 2016-01-25
Related to IP-518- close the db session properly

schimmy:
I realized that I wasn't closing the overall session properly, even
if I was closing the transaction. While this probably won't fix anything,
there's a decent change this will surface general connection issues via
the error passed back from the Close


bstein-clever:
LGTM

PR #51, opened 2016-01-29
Support manifest files

schimmy:
We need to support manifest files once the data gets above ~500 MB (compressed)
This adds manifest file support and gives more control over gzip, as well as makes some improvements to the tests.

### DIFF
--- a/cmd/s3_to_redshift.go
+++ b/cmd/s3_to_redshift.go
@@ -44,7 +44,7 @@ func fatalIfErr(err error, msg string) {
 	}
 }
 
-// in a transaction, truncate, create or update, and then copy from the s3 JSON file
+// in a transaction, truncate, create or update, and then copy from the s3 JSON file or manifest
 // yell loudly if there is anything different in the target table compared to config (different distkey, etc)
 func runCopy(db *redshift.Redshift, inputConf s3filepath.S3File, inputTable redshift.Table, targetTable *redshift.Table, truncate, gzip bool) error {
 	tx, err := db.Begin()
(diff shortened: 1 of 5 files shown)


cmd/s3_to_redshift.go
```diff
@@ -69,12 +70,8 @@ func runCopy(db *redshift.Redshift, inputConf s3filepath.S3File, inputTable reds
 	}
 
 	// COPY direct into it, ok to do since we're in a transaction
-	// assuming that we always want to copy from s3 and have gzip, so last 2 params are true
+	// assuming that we always want to copy from s3, so that parameter is always true
```
bstein-clever: Nit. Not actually a sentence, so this took me a while to parse.
I'm also not seeing any s3 parameters, so is this even an accurate comment, still?

bstein-clever:
Is the idea here that we'd manually switch tables over to using manifest files as they get to big/start failing? Or that we'd just always use them?
If the former, why support non-manifest files, and if the latter, it seems like we'll end up with extra failures/downtime.

bstein-clever:
Aside from a couple of logistical questions, LGTM.

cmd/s3_to_redshift.go
schimmy: good catch!

schimmy:
Great question! I think for all of our current use-cases we'll use manifest files (so we shouldn't need to worry about timeouts), but:
1) this is public so maybe others have started using that functionality
2) I like having raw json files as an option, as it's a little easier to set up that flow (mucking around with manifest files isn't super fun, and I'd like to have a low barrier to entry to start using the tool - when people run into size issues, there's an easy next step for them!)

PR #54, opened 2016-03-18
Add longtext column type

mb-dev:
adding longtext option, in cases where we need to store longer texts in redshift


schimmy:
LGTM!

PR #58, opened 2016-04-06
Up timeout

cgclever:


meliaj:
Concerned about what this means when we actually have redshift issues but let's try this out and tweak or set it back to 59 mins if it doesn't work out.

meliaj:
LGTM

PR #61, opened 2016-04-12
Increase timeout to 90m

meliaj:
Based on runs from last week and today we shouldn't need more than 90 mins to run the initial copy command and analyze command (and the first command may not be needed more than once!)
https://www.dropbox.com/s/f0n91susxgsqfih/Screenshot%202016-04-11%2017.23.59.png?dl=0


cgclever:
LGTM!

PR #62, opened 2016-04-22
IP-737: Replace matching data within time range

evaninja:
**JIRA**: https://clever.atlassian.net/browse/IP-737

**Overview**: We are updating key Pulse charts to refresh hourly (Instant Pulse!) and so as a result we'll need to modify some logic in s3-to-redshift to allow multiple syncs to run per day (and to refresh new data within a certain time range, specified by `granularity`). These changes:

1. Introduce a `granularity` flag, which determines how often we expect to store copies of data in each table. Currently, this is a daily interval for all tables involved.
2. Modify the `--force` conditions to allow updates to the most recent sync for new, incoming data (i.e. in the current day, or hour, depending on `granularity`).
3. Introduce an extra step in the SQL transaction (`TruncateInTimeRange`) that, for a particular sync, deletes data before the subsequent COPY refreshes that data.

**Testing**: Wrote unit tests for Redshift SQL functions. Performed manual testing against prod for the following cases:

- Verified that `--force` is no longer required for updates to the most recent data!
- New sync overwrites data, but does not append it. In essence, an idempotent operation! I tested this by deleting a subset of rows for a particular date, and ensuring that they were added back after running the s3-to-redshift sync (and that no duplicate rows were appended).
- Verified that `--force` is still required for updates outside of the most recent time granularity time range.
- Tested both `--granularity=day` and `--granularity=hour`

**Roll Out**:
- [x] Make sure to first modify `business-metrics` and `mongo-to-s3` to set `--granularity=day`. This is technically optional because this flag defaults to `day`, but would be nice for clarity purposes. At the very least we should document in those clients how often we expect the granularity to be.
- Optional: standardize the s3 filenames.. so that we truncate exactly to the desired granularity in the filename (Benji's suggestion).
- Reschedule `mongo-to-s3` and `business-metrics` cron jobs

### DIFF
--- a/s3_to_redshift.go
+++ b/s3_to_redshift.go
@@ -26,6 +26,7 @@ var (
 	dataDate        = flag.String("date", "", "data date we should process, must be full RFC3339")
 	configFile      = flag.String("config", "", "schema & table config to use in YAML format")
 	gzip            = flag.Bool("gzip", true, "whether target files are gzipped, defaults to true")
+	timeGranularity = flag.String("granularity", "day", "how often we expect to append new data")
 	// things which will would strongly suggest launching as a second worker are env vars
 	// also the secrets ... shhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh
 	host               = os.Getenv("REDSHIFT_HOST")
@@ -44,9 +45,31 @@ func fatalIfErr(err error, msg string) {
 	}
 }
 
+// helper function used for verifying inputs
+func getMapKeys(m map[string]bool) []string {
+	keys := make([]string, 0, len(m))
+	for k := range m {
+		keys = append(keys, k)
+	}
+	return keys
+}
+
+// Rounds down a dateTime to a granularity
+// For instance, 11:50AM will be truncated to 11:00AM
+// if given a granularity of an hour
+func truncateDate(date time.Time, granularity string) time.Time {
+	switch granularity {
+	case "hour":
+		return date.Truncate(time.Hour)
+	default:
+		// Round down to day granularity by default
+		return date.Truncate(24 * time.Hour)
+	}
+}
+
 // in a transaction, truncate, create or update, and then copy from the s3 JSON file or manifest
 // yell loudly if there is anything different in the target table compared to config (different distkey, etc)
-func runCopy(db *redshift.Redshift, inputConf s3filepath.S3File, inputTable redshift.Table, targetTable *redshift.Table, truncate, gzip bool) error {
+func runCopy(db *redshift.Redshift, inputConf s3filepath.S3File, inputTable redshift.Table, targetTable *redshift.Table, truncate, gzip bool, timeGranularity string) error {
 	tx, err := db.Begin()
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


s3_to_redshift.go
```diff
@@ -88,6 +107,14 @@ func runCopy(db *redshift.Redshift, inputConf s3filepath.S3File, inputTable reds
 func main() {
 	flag.Parse()
 
+	// verify that timeGranularity is a supported value. for convenience,
+	// we use the convention that granularities must be valid PostgreSQL dateparts
+	// (see: http://www.postgresql.org/docs/8.1/static/functions-datetime.html#FUNCTIONS-DATETIME-TRUNC)
+	supportedGranularities := map[string]bool{"hour": true, "day": true}
+	if !supportedGranularities[*timeGranularity] {
+		panic(fmt.Sprintf("Unsupported granularity, must be one of %v", reflect.ValueOf(supportedGranularities).MapKeys()))
```
cgclever: I'd strongly prefer we don't introduce `reflect` as a dependency, particularly when we only use it for getting keys of a map.

s3_to_redshift.go
```diff
@@ -44,9 +46,20 @@ func fatalIfErr(err error, msg string) {
 	}
 }
 
+// rounds down a dateTime to the timeGranularity
+func truncateDate(date time.Time) time.Time {
```
cgclever: nit, I'd prefer the time granularity be passed as an argument and not implicitly referred to by global variables.

s3_to_redshift.go
```diff
@@ -26,6 +27,7 @@ var (
 	dataDate        = flag.String("date", "", "data date we should process, must be full RFC3339")
 	configFile      = flag.String("config", "", "schema & table config to use in YAML format")
 	gzip            = flag.Bool("gzip", true, "whether target files are gzipped, defaults to true")
+	timeGranularity = flag.String("granularity", "day", "how often we expect to append new data")
```
cgclever: I wonder if time granularity should be a required flag rather than defaulting to a day. Do what you want here, but I'd recommend thinking about it.

s3_to_redshift.go
cgclever: Also when you say "rounds down..." does it literally round down? For example does 11:50 round to 11:00 or 12:00?

cgclever:
A few suggestions.

s3_to_redshift.go
evaninja: Good points. Yup, the desired behavior is to truncate. Added additional parameter + comments in f69d70e

s3_to_redshift.go
evaninja: Valid. Removed in 9bcaf5c

s3_to_redshift.go
evaninja: Great point. I think that since there only exist use cases for day granularity at the moment, it's OK to default to day, but perhaps as more granularities are required we can revisit this question

evaninja:
Thanks for the review! Added a few fixups - PTAL

s3_to_redshift.go
cgclever: Even in your PR you mention explicitly referencing a day in the payload. Feeling like if we're doing that anyways, let's be strict about requiring you pass in the granularity.

evaninja: How strongly do you feel about this? I have a preference to keep the defaults behavior, given that no use case actually exists right now for non-day granularity (and I generally prefer the principle of engineering things as the use case arises). However, if you feel strongly that we should enforce day granularity, we can absolutely add it in.

cgclever: I feel strongly enough to bring it up twice, but it's your call. If you feel we should have day as the default then let's keep it as the default.

evaninja: For record-keeping: Christian and I chatted about this in person, and agreed that for now we ought to keep the default behavior (since changing it would perform a backwards incompatible change). However, he did bring up the point that this would be a potential stumbling block for users (who might not expect to know how to use `granularity`). As a result, documentation has been added in dd0fdc4

cgclever:
LGTM!

PR #65, opened 2016-05-16
increasing size of longtext to be really long

nikhilpandit:
We need a longer `longtext` for some of the use cases we are considering. Discussed with @schimmy and he recommended modifying the existing `longtext` instead of adding another one, since `longtext` is not widely used right now.


bstein-clever:
LGTM!

PR #69, opened 2016-07-22
Ip 847 allow csv with delimiters

schimmy:
**Overview:**
We want to support CSV files for our `COPY` commands, specifically because we need to ingest data which was `UPLOAD`ed, and that command only supports CSV

**Testing:**
Tested locally, having trouble testing that the JSON stuff still works however

**Roll Out:**
Simple rollout as it's idempotent

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ The `schema`, `tables`, and `bucket` flags also are critical. `granularity` is r
 - `force`: refresh the data even if the data date is after the current `s3` input date
 - `date`:  the date string for the data in question
 - `config`: override of the usual auto-discovery of the config
+- `delimiter`: required to use CSV files, what the file is delimited in (likely use the '|' pipe character as that is AWS' default)
 - `granularity`: how often we expect to append new data for each table (i.e. daily, or hourly buckets)
 
 #### Note on general usage:
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ The `schema`, `tables`, and `bucket` flags also are critical. `granularity` is r
 - `force`: refresh the data even if the data date is after the current `s3` input date
 - `date`:  the date string for the data in question
 - `config`: override of the usual auto-discovery of the config
-- `delimiter`: required to use CSV files, what the file is delimited in (likely use the '|' pipe character as that is AWS' default)
+- `delimiter`: required to use CSV files, what the file is delimited in (likely use the '|' pipe character as that is AWS' default). If `""` then JSON copy is assumed
 - `granularity`: how often we expect to append new data for each table (i.e. daily, or hourly buckets)
 
 #### Note on general usage:
@@ -121,7 +121,7 @@ Currently supported granularities are `hour` and `day`.
 Assuming that environment variables have been set:
 ```
 go run cmd/s3_to_redshift.go -schema=api_hits -tables=pages,sessions \
-  -bucket=analytics -config=s3://analytics/api.yml -date=2015-07-01T00:00:00Z -force=true
+  -bucket=analytics -config=s3://analytics/api.yml -date=2015-07-01T00:00:00Z -force=true -delimiter="|"
 ```
 
 ## Vendoring
--- a/redshift/redshift.go
+++ b/redshift/redshift.go
@@ -315,11 +315,11 @@ func (r *Redshift) UpdateTable(tx *sql.Tx, targetTable, inputTable Table) error 
 	return nil
 }
 
-// Copy copies JSON data present in an S3 file into a redshift table.
-// It also supports JSON data pointed at by a manifest file, if you pass in a manifest file.
+// Copy copies either CSV or JSON data present in an S3 file into a redshift table.
+// It also supports CSV or JSON data pointed at by a manifest file, if you pass in a manifest file.
 // this is meant to be run in a transaction, so the first arg must be a sql.Tx
 // if not using jsonPaths, set s3File.JSONPaths to "auto"
-func (r *Redshift) Copy(tx *sql.Tx, f s3filepath.S3File, creds, gzip bool) error {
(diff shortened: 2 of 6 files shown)


s3_to_redshift.go
```diff
@@ -99,8 +100,11 @@ func runCopy(db *redshift.Redshift, inputConf s3filepath.S3File, inputTable reds
 	}
 
 	// COPY direct into it, ok to do since we're in a transaction
-	if err = db.JSONCopy(tx, inputConf, true, gzip); err != nil {
-		return fmt.Errorf("err running JSON copy: %s", err)
+	// can't switch on file ending as manifest files b/c
+	// manifest files obscure the underlying file types
+	// instead just pass the delimiter along even if it's null
+	if err = db.Copy(tx, inputConf, delimiter, true, gzip); err != nil {
```
cgclever: Nit, it would be nice if these error variables were scoped and not reused.
```
if err := db.Copy...
```
or even
```
if copyErr :=
```

README.md
```diff
@@ -61,6 +61,7 @@ The `schema`, `tables`, and `bucket` flags also are critical. `granularity` is r
 - `force`: refresh the data even if the data date is after the current `s3` input date
 - `date`:  the date string for the data in question
 - `config`: override of the usual auto-discovery of the config
+- `delimiter`: required to use CSV files, what the file is delimited in (likely use the '|' pipe character as that is AWS' default)
```
evaninja: nit: Are we recommending '|' or '\t'?

redshift/redshift.go
```diff
@@ -334,7 +334,19 @@ func (r *Redshift) JSONCopy(tx *sql.Tx, f s3filepath.S3File, creds, gzip bool) e
 	if f.Suffix == "manifest" {
 		manifestSQL = "manifest"
 	}
-	copySQL := fmt.Sprintf(`COPY "%s"."%s" FROM '%s' WITH %s JSON '%s' REGION '%s' TIMEFORMAT 'auto' TRUNCATECOLUMNS STATUPDATE ON COMPUPDATE ON %s %s`, f.Schema, f.Table, f.GetDataFilename(), gzipSQL, f.JSONPaths, f.Bucket.Region, manifestSQL, credSQL)
+
+	// default to CSV
+	jsonSQL := ""
+	jsonPathsSQL := ""
+	delimSQL := fmt.Sprintf("DELIMITER AS '%s' REMOVEQUOTES TRIMBLANKS EMPTYASNULL ACCEPTANYDATE", delimiter) // always removequotes, UNLOAD should add quotes
```
evaninja: 👍 to this comment

README.md
schimmy: hmm, I think, if we can get away with it, we should go for the pipe since that is Amazon's default and a bit easier to work with as humans generally. I'm hoping that the issue we saw really was a pathological case...

But I haven't thought too much about it, let's talk it over!

redshift/redshift.go
```diff
@@ -334,7 +334,19 @@ func (r *Redshift) JSONCopy(tx *sql.Tx, f s3filepath.S3File, creds, gzip bool) e
...
+	// figure out if we're doing JSON - no delim means JSON
+	if delimiter == "" {
+		jsonSQL = "JSON"
+		jsonPathsSQL = "'auto'"
```
evaninja: Cool that we were able to get rid of S3File.JSONPaths!

s3filepath/s3filepath.go
```diff
@@ -69,21 +68,18 @@ func CreateS3File(pc PathChecker, bucket S3Bucket, schema, table, suppliedConf s
...
-	if pc.FileExists(inputFile.GetDataFilename()) {
-		return &inputFile, nil
+	// Try to find manifest or data files out of the following patterns, in order
+	// we try to get in order as otherwise
+	for _, suffix := range []string{
+		"manifest", // 1) manifest file
+		"json.gz",  // 2) gzipped json file
+		"json",     // 3) json file
+		".gz",      // 4) gzipped csv file (.gz)
+		""} {       // 5) csv file (no suffix :-/)
```
evaninja: No suffix here because of the output of UNLOAD?

s3_to_redshift.go
```diff
@@ -67,9 +68,9 @@ func truncateDate(date time.Time, granularity string) time.Time {
 	}
 }
 
-// in a transaction, truncate, create or update, and then copy from the s3 JSON file or manifest
+// in a transaction, truncate, create or update, and then copy from the s3 dagta file or manifest
```
evaninja: what's dagta?

README.md
evaninja: As an additional suggestion I would also clarify here that if delimiter is `""` then we'll use a JSON copy!

redshift/redshift_test.go
```diff
@@ -477,6 +473,100 @@ func TestTruncate(t *testing.T) {
...
+	}
+	// test with creds and GZIP
+	sql := `COPY "%s"."%s" FROM '%s' WITH %s REGION '%s' TIMEFORMAT 'auto' TRUNCATECOLUMNS STATUPDATE ON COMPUPDATE ON CREDENTIALS 'aws_access_key_id=%s;aws_secret_access_key=%s' DELIMITER AS '|' EMPTYASNULL ACCEPTANYDATE`
+	execRegex := fmt.Sprintf(sql, schema, table, s3File.GetDataFilename(),
+		"GZIP", region, accessID, secretKey)
+
+	db, mock, err := sqlmock.New()
+	assert.NoError(t, err)
+	defer db.Close()
+	mockrs := Redshift{db}
```
evaninja: nit: would rename to mockRs

evaninja:
LGTM

Context: chatted with Schimmy offline, we decided that `DELIMITER AS '|' ESCAPE REMOVEQUOTES` would be sufficient to adequately encode our data.

As a result, UNLOAD will need `DELIMITER AS '|' ESCAPE ADDQUOTES`

PR #70, opened 2016-08-17
Find Region From Bucket

cgclever:
This dynamically determines the AWS region instead of requiring it as an input.

Tested by running locally and ensuring the correct region is selected.

### DIFF
--- a/s3_to_redshift.go
+++ b/s3_to_redshift.go
@@ -10,6 +10,9 @@ import (
 	"strings"
 	"time"
 
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
 	env "github.com/segmentio/go-env"
 
 	redshift "github.com/Clever/s3-to-redshift/redshift"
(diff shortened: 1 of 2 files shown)


s3_to_redshift.go
```diff
@@ -113,6 +114,12 @@ func runCopy(db *redshift.Redshift, inputConf s3filepath.S3File, inputTable reds
 	return nil
 }
 
+func init() {
+	if *inputRegion != "" {
+		awsRegion = *inputRegion
```
schimmy: If we set this via command line argument, we should remove the environment variable here - otherwise it's confusing to a user which one is actually controlling the behavior

s3_to_redshift.go
```diff
@@ -113,6 +114,12 @@ func runCopy(db *redshift.Redshift, inputConf s3filepath.S3File, inputTable reds
 	return nil
 }
 
+func init() {
```
schimmy: why set this in an init function? I'd suggest doing it in the same place as the other args, after flag.parse

schimmy:
Sweet, if this works, it's WAY cleaner :)

schimmy:
LGTM

PR #71, opened 2016-08-18
Add back in Region requirement in launch

cgclever:
We want to keep this required in case we need to roll back the change made [here](https://github.com/Clever/s3-to-redshift/pull/70).


bstein-clever:
LGTM

PR #73, opened 2016-08-25
Bump gearcmd version

evaninja:
**JIRA**: https://clever.atlassian.net/browse/IP-1002 (oncall)

**Overview**: Update `s3-to-redshift` to the latest version of `gearcmd`. This will be useful for `heartbeat` logging, which can notify us when long-running jobs are running so that we can debug live.

**Testing**: See: https://github.com/Clever/gearcmd/pull/50

**Roll Out:**

Remember to revert the .drone.yml change

CC: Assigning you @bstein-clever since Schimmy is busy AFAICT

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ RUN apt-get update -y && \
     apt-get update -y && \
     apt-get install -y curl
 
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -20,7 +20,7 @@ publish:
     - $(git rev-parse --short HEAD)
     username: $$docker_username
     when:
-      branch: master
```
bstein-clever: Obligatory "revert this" comment.

bstein-clever:
LGTM

evaninja:
@bstein-clever thanks for the quick review

PR #75, opened 2016-09-12
Include WORKER_NAME for publishing to gearman

cgclever:
**[Associated SALT PR](https://github.com/Clever/salt/pull/2193)**


cgclever:
Verbal LGTM from @evaninja

PR #76, opened 2016-09-12
Fix Dockerfile cmd

cgclever:
Fix typo


cgclever:
LGTM from @jiancheung

PR #80, opened 2016-11-23
Add clone for elb logs cluster

ltclever:
**JIRA**:
none

**Overview**:
the other side of https://github.com/Clever/record-keeper/pull/20

**Testing**:

**Roll Out**:
- [ ] Remember to deploy both `s3-to-redshift` and `s3-to-redshift-fast` (and the new clone!)


cgclever:
LGTM!

PR #82, opened 2016-12-15
IP-1211: Log When Job Finishes

evaninja:
**JIRA**: https://clever.atlassian.net/browse/IP-1211

**Overview**: Uses log-routing to log when a job finishes to SignalFx, along with the payload and pass/fail. Also imports kayvee v6 and updates our usage of Errorf to meet new linting standards.

**Testing**: Wrote unit test for new logger package + ran locally.

**Roll Out**:

CC @xavi-


logger/logger.go
```diff
@@ -0,0 +1,77 @@
+package logger
+
+import (
+	"gopkg.in/Clever/kayvee-go.v6/logger"
+)
+
+var log logger.KayveeLogger
```
bstein-clever: I really wish there was a way so we didn't have to do this boilerplate.

bstein-clever:
A couple of nits, otherwise LGTM

kvconfig.yml
```diff
@@ -0,0 +1,10 @@
+routes:
+  job-finished:
+    matchers:
+      title: [ "job-finished" ]
+    output:
+      type: "alerts"
+      series: "job.finished"
+      dimensions: [ "function", "payload" ]
+      value: "value"
+      stat_type: "gauge"
```
xavi-: I'm assuming that "value" has a the delay value?

logger/logger.go
xavi-: Creating your new logger shouldn't be required.  It should be okay to use `logger.KayveeLogger` directly.

kvconfig.yml
evaninja: We log a value of 1 if the job completed successfully, 0 otherwise

xavi-: Making it a `counter` might make more sense

main.go
```diff
@@ -6,15 +6,18 @@ import (
...
+	"path"
 	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/kardianos/osext"
 	env "github.com/segmentio/go-env"
+	kvlogger "gopkg.in/Clever/kayvee-go.v6/logger"
```
evaninja: 👍

kvconfig.yml
evaninja: Talked with Xavi offline - we are shooting for a gauge

PR #83, opened 2016-12-16
Fix Dockerfile issue

evaninja:
I copied the kvconfig.yml to the wrong place :(

*Roll Out*:
- [ ] Remember to deploy both s3-to-redshift and s3-to-redshift-fast


ahsojar:
LGTM

PR #85, opened 2017-01-11
Route worker name correctly

evaninja:
**JIRA**: Oncall, N/A

**Overview**: Previously we weren't routing the worker name to SignalFx correctly since `function` appears to now be logged as `source` by Kayvee. We want to be able to alert on both payload and worker name!

**Testing**: Verified that the routed logs now contain the worker name.

**Roll Out**:


cgclever:
LGTM!

PR #91, opened 2017-03-25
alternate: update aws-sdk-go

mohit:
**Overview**:

Upgrade `aws-sdk-go` to the latest release. This will be required for moving to roles, to support redshift roles and for integration with AWS Batch. 

**Testing**:

none yet - looking for some help here. 

**Roll Out**:


Godeps/Godeps.json
```diff
@@ -11,7 +11,7 @@
 	"Deps": [
 		{
 			"ImportPath": "github.com/Clever/pathio",
-			"Comment": "v3.0.0-2-gc9ab5fa",
+			"Comment": "v2.0.0-23-gc9ab5fa",
```
schimmy: huh- so we had to downgrade a major version of pathio - will this cause a problem? what changed in Pathio?

mohit: seems like this is just a comment. The Rev is still the same. Honestly don't really know wo this change got here.

PR #92, opened 2017-04-26
Truncate like we mean it

evaninja:
**JIRA**: https://clever.atlassian.net/browse/IPOC-268

**Overview**: Truncate is necessary since it won't require a VACUUM afterwards. (`DELETE FROM` leaves "deleted rows" lying around in Redshift which adversely affects query performance)

**Testing**:

**Roll Out**:

### DIFF
--- a/redshift/redshift.go
+++ b/redshift/redshift.go
@@ -358,9 +358,8 @@ func (r *Redshift) Copy(tx *sql.Tx, f s3filepath.S3File, delimiter string, creds
 }
 
 // Truncate deletes all items from a table, given a transaction, a schema string and a table name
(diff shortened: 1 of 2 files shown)


redshift/redshift.go
```diff
@@ -360,7 +360,7 @@ func (r *Redshift) Copy(tx *sql.Tx, f s3filepath.S3File, delimiter string, creds
 // Truncate deletes all items from a table, given a transaction, a schema string and a table name
 // you should run vacuum and analyze soon after doing this for performance reasons
```
schimmy: you should remove this line as well

PR #94, opened 2017-04-28
Revert "Truncate like we mean it"

ahsojar:
Reverts Clever/s3-to-redshift#92, until we decide if "Truncate" is the right move here


ahsojar:
verbal LGTM from Evan

PR #96, opened 2017-05-02
Vacuum / analyze after truncating

kvigen:
This PR does two things:
1. Handles an empty table in this worker
2. Runs vacuum / analyze on the table if we truncated it. We could be a bit more aggressive and do these even when we don't truncate the whole table, but I figure it's better to start conservative.

Theoretically after we merge this we could remove some of the redshift-respawn jobs that vacuum tables like mongo.students.

Tested by running locally against the dev cluster.

**JIRA**:

**Overview**:

**Testing**:

**Roll Out**:

### DIFF
--- a/main.go
+++ b/main.go
@@ -155,9 +155,8 @@ func runCopy(db *redshift.Redshift, inputConf s3filepath.S3File, inputTable reds
 	}
 
 	if truncate {
-		// If we've truncated the table we should run vacuum and analyze to sort the
-		// data, remove the deleted rows, and update the statistics.
-		if err := db.VacuumAnalyze(inputConf.Schema, inputTable.Name); err != nil {
+		// If we've truncated the table we should run vacuum to clear out the old data
+		if err := db.VacuumDelete(inputConf.Schema, inputTable.Name); err != nil {
 			return fmt.Errorf("err vacuuming or analyzing database: %s", err)
 		}
 	}
--- a/main.go
+++ b/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"database/sql"
 	"flag"
 	"fmt"
 	"log"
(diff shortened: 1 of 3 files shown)


redshift/redshift.go
```diff
@@ -360,6 +363,8 @@ func (r *Redshift) Copy(tx *sql.Tx, f s3filepath.S3File, delimiter string, creds
 // Truncate deletes all items from a table, given a transaction, a schema string and a table name
 // you should run vacuum and analyze soon after doing this for performance reasons
 func (r *Redshift) Truncate(tx *sql.Tx, schema, table string) error {
+	// We run 'DELETE FROM' instead of 'TRUNCATE' because 'TRUNCATE' can't be run in a transaction.
```
bstein-clever: Thank you for adding this! Now we're less likely to do it again. :)

main.go
```diff
@@ -154,6 +153,14 @@ func runCopy(db *redshift.Redshift, inputConf s3filepath.S3File, inputTable reds
 	if err := tx.Commit(); err != nil {
 		return fmt.Errorf("err committing transaction: %s", err)
 	}
+
+	if truncate {
+		// If we've truncated the table we should run vacuum and analyze to sort the
+		// data, remove the deleted rows, and update the statistics.
+		if err := db.VacuumAnalyze(inputConf.Schema, inputTable.Name); err != nil {
```
bstein-clever: Do you think it's worth limiting this to just a `VACUUM DELETE ONLY`, for efficiency http://docs.aws.amazon.com/redshift/latest/dg/r_VACUUM_command.html ?

We could then run a full analyze later on.

kvigen: I'm not sure how expensive the other parts are, but if we're worried about performance i'm happy to start with just `DELETE`

bstein-clever: Sounds like a plan. I'd rather start small and work bigger - then we know which changes impacted performance.

kvigen: 👍

PR #97, opened 2017-05-04
Fix new table creation

evaninja:
**JIRA**: N/A, quick oncall fix

**Overview**: A recent change to `s3-to-redshift` touched some logic that prevented new tables from being created (https://github.com/Clever/s3-to-redshift/commit/5d359adc771d75c60aaac19a46ab0959a98d4d68#diff-7ddfb3e035b42cd70649cc33393fe32cR212).

This updates the logic to allow for new table creation, by short circuiting `GetTableMetadata` with no error if the table does not exist - which is valid!

**Testing**:
- [x] Updated automated test

**Roll Out**:


bstein-clever:
This looks good, but can we add a few comments about the flows/checks?
I feel like if we'd explicitly stated why we made the branches/calls/returns we did to explain the various cases, we wouldn't have stepped on ourselves so many times recently.

PR #98, opened 2017-05-05
Don't vacuum here, call out to redshift-vacuum

bstein-clever:
**JIRA**:
https://clever.atlassian.net/browse/IPOC-283

**Overview**:
Redshift vacuum jobs need to be queued, so we'll trigger them in their own gearman worker, rather than doing them here.

**Testing**:
We'll do it in prod! (As we roll out.)

**Roll Out**:
It's complicated. See plan here: https://github.com/Clever/redshift-vacuum/pull/1


main.go
```diff
@@ -156,8 +180,24 @@ func runCopy(db *redshift.Redshift, inputConf s3filepath.S3File, inputTable reds
...
+			endpoint := gearmanAdminURL + fmt.Sprintf("/%s", vacuumWorker)
+			payload := fmt.Sprintf(`--delete %s."%s"`, inputConf.Schema, inputTable.Name)
+			req, err := http.NewRequest("POST", endpoint, bytes.NewReader([]byte(payload)))
+			if err != nil {
+				log.Fatalf("Error creating new request: %s", err)
+			}
+			req.Header.Add("Content-Type", "text/plain")
+			_, err = client.Do(req)
+			if err != nil {
+				log.Fatalf("Error submitting job:%s", err)
```
evaninja: nit: missing space between `:`and `%s`

bstein-clever: Good eye, thanks!

PR #100, opened 2017-05-09
Handle column reorder

kvigen:
For now that means erroring with a clear message, so we don't accidentally screw things up, and so that we can fix the issue quickly.

Since view-materializer + s3-to-redshift together assume that the columns are the same (and in the same order) in both the source and the destination.  If the source just has new columns we add them. Otherwise we note the errors and fail.

This change also makes us continue after a single failure, so one bad table config doesn't stop us from loading any data.

**JIRA**:

**Overview**:

**Testing**:

**Roll Out**:

### DIFF
--- a/main.go
+++ b/main.go
@@ -239,6 +239,7 @@ func main() {
 	db, err := redshift.NewRedshift(host, port, dbName, user, pwd, timeout)
 	fatalIfErr(err, "error getting redshift instance")
 
+	var tablesWithErrors []string
 	// for each table passed in - likely we could goroutine this out
 	for _, t := range strings.Split(*inputTables, ",") {
 		log.Printf("attempting to run on schema: %s table: %s", *inputSchemaName, t)
@@ -272,9 +273,16 @@ func main() {
 			log.Printf("Forcing update of inputTable: %s", inputConf.Table)
 		}
 
-		fatalIfErr(runCopy(db, *inputConf, *inputTable, targetTable, *truncate, *gzip, *delimiter, *timeGranularity), "Issue running copy")
-		// DON'T NEED TO CREATE VIEWS - will be handled by the refresh script
-		log.Printf("done with table: %s.%s", inputConf.Schema, t)
+		if err := runCopy(db, *inputConf, *inputTable, targetTable, *truncate, *gzip, *delimiter, *timeGranularity); err != nil {
+			log.Printf("error running copy for table %s: %s", t, err)
+			tablesWithErrors = append(tablesWithErrors, t)
+		} else {
+			// DON'T NEED TO CREATE VIEWS - will be handled by the refresh script
+			log.Printf("done with table: %s.%s", inputConf.Schema, t)
+		}
+	}
+	if len(tablesWithErrors) > 0 {
+		log.Fatalf("error loading tables: %+v\n", tablesWithErrors)
 	}
 	log.Println("done with full run")
 }
--- a/main.go
+++ b/main.go
@@ -21,6 +21,7 @@ import (
 	"github.com/Clever/s3-to-redshift/logger"
 	redshift "github.com/Clever/s3-to-redshift/redshift"
 	s3filepath "github.com/Clever/s3-to-redshift/s3filepath"
+	"github.com/hashicorp/go-multierror"
 )
 
 var (
@@ -239,7 +240,7 @@ func main() {
 	db, err := redshift.NewRedshift(host, port, dbName, user, pwd, timeout)
 	fatalIfErr(err, "error getting redshift instance")
 
-	var tablesWithErrors []string
+	var copyErrors error
 	// for each table passed in - likely we could goroutine this out
 	for _, t := range strings.Split(*inputTables, ",") {
 		log.Printf("attempting to run on schema: %s table: %s", *inputSchemaName, t)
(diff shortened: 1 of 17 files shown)


main.go
```diff
@@ -272,9 +273,16 @@ func main() {
 			log.Printf("Forcing update of inputTable: %s", inputConf.Table)
 		}
 
-		fatalIfErr(runCopy(db, *inputConf, *inputTable, targetTable, *truncate, *gzip, *delimiter, *timeGranularity), "Issue running copy")
-		// DON'T NEED TO CREATE VIEWS - will be handled by the refresh script
-		log.Printf("done with table: %s.%s", inputConf.Schema, t)
+		if err := runCopy(db, *inputConf, *inputTable, targetTable, *truncate, *gzip, *delimiter, *timeGranularity); err != nil {
```
bstein-clever: I was actually a big fan of https://github.com/hashicorp/go-multierror when I last needed something similar to this. It might make it a little simpler, but your call.

redshift/redshift.go
```diff
@@ -310,6 +310,9 @@ func checkSchemas(targetTable, inputTable Table) ([]string, []error) {
 
 		targetCol := targetTable.Columns[idx]
 		mismatchedTemplate := "mismatched column: %s property: %s, input: %v, target: %v"
+		if inCol.Name != targetCol.Name {
+			errors = append(errors, fmt.Errorf(mismatchedTemplate, inCol.Name, "Nane", inCol.Name, targetCol.Name))
```
bstein-clever: Nit: Type in `"Nane"`

redshift/redshift.go
```diff
@@ -318,12 +282,56 @@ func (r *Redshift) UpdateTable(tx *sql.Tx, targetTable, inputTable Table) error
 		log.Printf("Running command: %s", op)
 		_, err = alterStmt.Exec()
 		if err != nil {
-			return err
+			return fmt.Errorf("issue running statement %s: %s", op, err)
```
bstein-clever: Nice, thanks.

redshift/redshift_test.go
```diff
@@ -591,3 +465,92 @@ func TestTruncateInTimeRange(t *testing.T) {
...
+		ColInfo{Name: "DateColumn", PrimaryKey: false},
+		ColInfo{Name: "IntColumn2", PrimaryKey: true},
+	}}
+	columnOps, errors := checkSchemas(t1, t2)
+	assert.Equal(t, 0, len(errors), fmt.Sprintf("errors: %+v\n", errors))
+	assert.Equal(t, 2, len(columnOps))
+}
+
+func TestCheckSchemasDiffs(t *testing.T) {
+	// Do a re-order and a type difference
```
bstein-clever: Should we test these two individually, to make sure that both trigger errors?
We've definitely seen ones before where there's a reorder, but all the values are strings, so the load happens cleanly, so "appears" to be OK - but the data has become corrupted downstream. I realize that string->string is too hard for us to catch, currently, but still a thought.

redshift/redshift.go
```diff
@@ -318,12 +282,56 @@ func (r *Redshift) UpdateTable(tx *sql.Tx, targetTable, inputTable Table) error
...
 		}
 	}
 	return nil
 }
 
+// checkSchemas takes in two tables and compares their column schemas to make sure they're compatible.
+// If they have any mismatched columns they are returned in the errors array. If the input table has
+// columns at the end that the target table does not then the appropriate alert tables sql commands are
```
bstein-clever: Nit: `s/alert/alter/`

redshift/redshift.go
```diff
@@ -318,12 +282,56 @@ func (r *Redshift) UpdateTable(tx *sql.Tx, targetTable, inputTable Table) error
...
+	if len(inputTable.Columns) < len(targetTable.Columns) {
+		errors = append(errors, fmt.Errorf("target table has more columns than the input table"))
+	}
+
+	for idx, inCol := range inputTable.Columns {
+		if len(targetTable.Columns) <= idx {
+			fmt.Printf("Missing column -- running alter table\n")
+			alterSQL := fmt.Sprintf(`ALTER TABLE "%s"."%s" ADD COLUMN %s`, targetTable.Meta.Schema, targetTable.Name, getColumnSQL(inCol))
+			columnOps = append(columnOps, alterSQL)
+			continue
```
bstein-clever: Technically, I think we should have a targetTable.Columns++ here, but it's probably fine.

redshift/redshift.go
```diff
@@ -318,12 +282,56 @@ func (r *Redshift) UpdateTable(tx *sql.Tx, targetTable, inputTable Table) error
...
 	return nil
 }
 
+// checkSchemas takes in two tables and compares their column schemas to make sure they're compatible.
+// If they have any mismatched columns they are returned in the errors array. If the input table has
+// columns at the end that the target table does not then the appropriate alert tables sql commands are
+// returned.
+func checkSchemas(targetTable, inputTable Table) ([]string, []error) {
+	var columnOps []string
+	var errors []error
```
bstein-clever: Same suggestion about multierror, still optional.

main.go
kvigen: Ah, that's cool. Will give it a try

redshift/redshift_test.go
kvigen: I've added some extra tests and check the number of errors returned. Also, with the current approach we can determine if string -> string errors occurred because we check that the names of the columns match.

PR #102, opened 2017-05-23
Don't check for column ordering if schema is mongo

kvigen:
The input to the Mongo schema is json files, so the ordering doesn't
actually matter.

**JIRA**:

**Overview**:

**Testing**:

**Roll Out**:

### DIFF
--- a/redshift/redshift.go
+++ b/redshift/redshift.go
@@ -294,9 +294,19 @@ func (r *Redshift) UpdateTable(tx *sql.Tx, targetTable, inputTable Table) error 
 // columns at the end that the target table does not then the appropriate alter tables sql commands are
 // returned.
 func checkSchemas(targetTable, inputTable Table) ([]string, error) {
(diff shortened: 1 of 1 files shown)


redshift/redshift.go
```diff
@@ -294,9 +294,19 @@ func (r *Redshift) UpdateTable(tx *sql.Tx, targetTable, inputTable Table) error
 // columns at the end that the target table does not then the appropriate alter tables sql commands are
 // returned.
 func checkSchemas(targetTable, inputTable Table) ([]string, error) {
+	// If the schema is mongo then we know the input files are json so ordering doesn't matter. At
+	// some point we could handle this in a more general way by checking if the input files are json.
+	// This wouldn't be too hard, but we would have to peak in the manifest file to check if all the
+	// files references are json.
```
ahsojar: thanks for adding this comment!

PR #104, opened 2017-07-11
Add support for streaming payloads

kvigen:
Now that we support granularity = 'stream' (see https://github.com/Clever/record-keeper/pull/28) we need to support making that idempotent. That means adding start / end dates instead of just granularities to the payload. That way we can delete all the day in the time range before adding the new rows.

**JIRA**:

**Overview**:

**Testing**:

**Roll Out**:

### DIFF
--- a/main.go
+++ b/main.go
@@ -36,6 +36,8 @@ var (
 	gzip            = flag.Bool("gzip", true, "whether target files are gzipped, defaults to true")
 	delimiter       = flag.String("delimiter", "", "delimiter for CSV files, usually pipe character. If empty then JSON will be assumed.")
 	timeGranularity = flag.String("granularity", "day", "how often we expect to append new data")
+	streamStart     = flag.String("streamStart", "", "The start of the streamed data. Only used if granularity='stream'. Used to ensure idempotency")
+	streamEnd       = flag.String("streamEnd", "", "The end of the streamed data. Only used if granularity='stream'. Used to ensure idempotency")
 	// things which will would strongly suggest launching as a second worker are env vars
 	// also the secrets ... shhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh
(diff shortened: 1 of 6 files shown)


main.go
```diff
@@ -204,12 +211,34 @@ func runCopy(db *redshift.Redshift, inputConf s3filepath.S3File, inputTable reds
 	return nil
 }
 
+func startEndFromGranularity(t time.Time, granularity string) (time.Time, time.Time) {
+	var duration time.Duration
+	if granularity == "day" {
+		duration = time.Hour * 24
+	} else {
+		duration = time.Hour * 24 * 7
```
ahsojar: perhaps we should hardcode this for `granularity == "hour"` in case we add other granularities in the future? Otherwise it will always default to this

main.go
```diff
@@ -204,12 +211,34 @@ func runCopy(db *redshift.Redshift, inputConf s3filepath.S3File, inputTable reds
...
 // If the destination table does not exist, the worker creates it
 // If the destination table lacks columns, the worker creates those as well
 // The worker also uses a column in the data to figure out whether the s3 data is
 // newer than what already exists.
 func main() {
+	dir, err := osext.ExecutableFolder()
+	if err != nil {
+		log.Fatal(err)
+	}
+	err = logger.SetGlobalRouting(path.Join(dir, "kvconfig.yml"))
```
ahsojar: thank you for adding this!

main.go
```diff
@@ -152,11 +151,11 @@ func runCopy(db *redshift.Redshift, inputConf s3filepath.S3File, inputTable reds
 		var start, end time.Time
 		var err error
 		if timeGranularity == "stream" {
-			start, err = time.Parse(*streamStart, "2006-01-02T15:04:05.000Z")
+			start, err = time.Parse("2006-01-02T15:04:05", *streamStart)
 			if err != nil {
 				return err
 			}
-			end, err = time.Parse(*streamEnd, "2006-01-02T15:04:05.000Z")
```
kvigen: :man_facepalming:

PR #105, opened 2017-07-24
Allow ignore of sortkey and distkey

schimmy:
**JIRA**:
https://clever.atlassian.net/browse/IP-1532

**Overview**:
If you want to pull from a view and materialize into a table, for instance, you won't get a sort or distkey. This change enforces creating a table with a sort & distkey, and also relaxes the constraint that the destination table must have the same sortkey & distkey _as long as the source table has no sort or distkey_


schimmy:
cc @kvigen and @ahsojar as we were talking about this

PR #106, opened 2017-08-02
Escape quotes in argument to redshift-vacuum

evaninja:
**JIRA**: N/A, oncall.

**Overview**: Turns out we weren't actually surrounding our table names with double quotes after all - Golang's os.Args requires double quotes to be escaped. (A simple way to verify this is to write a dummy program and test out how it parses your string arguments by using `flags.Parse()` or printing out `os.Args[1:]`).

**Testing**: Discovered as part of https://github.com/Clever/redshift-vacuum/pull/8

**Roll Out**:


bstein-clever:
Assuming tested and working, LGTM. Nice find!

PR #107, opened 2017-08-10
Use IAM roles instead of users for redshift

ahsojar:
JIRA:
https://clever.atlassian.net/browse/IP-1544

cc @alsmola

Overview:
Use IAM roles instead of user creds.

Ark config: https://github.com/Clever/ark-config/pull/538

Testing:
Tested locally, I will make sure prod permissions also works fine once i deploy.


Makefile
```diff
@@ -20,8 +20,8 @@ export SERVICE_GEARMAN_ADMIN_HTTP_HOST ?= x
 export SERVICE_GEARMAN_ADMIN_HTTP_PORT ?= x
 export SERVICE_GEARMAN_ADMIN_HTTP_PROTO ?= x
 export AWS_REGION ?= x
-export AWS_ACCESS_KEY_ID ?= x
-export AWS_SECRET_ACCESS_KEY ?= x
```
alsmola: Stray newline?

PR #108, opened 2017-08-16
IP-1556: Add support for EOD in PST

evaninja:
Please view commit by commit

**JIRA**: https://clever.atlassian.net/browse/IP-1551

**Overview**: Extends s3-to-redshift to allow treating Redshift timestamps (without timezone) as PST timestamps. This ultimately allows us to implement EOD as midnight PST for historical tables by simply changing the data date column to `_record_timestamp_in_pst` for non-streaming `record-keeper` jobs (to be implemented).

**Testing**:
- [x] Wrote tests for new functionality
- [ ] Test in staging

**Roll Out**:

### DIFF
--- a/main.go
+++ b/main.go
@@ -106,6 +106,23 @@ func truncateDate(date time.Time, granularity string) time.Time {
 	}
 }
 
+// Calculates whether or not input data (s3) is more stale than target data (Redshift)
+// Expects:
+// - inputDataDate corresponds to the s3 data timestamp of the job
+// - targetDataDate is the maximum timestamp of the DB table
+// - granularity indicating how often data snapshots are recorded in the target
+func isInputDataOlder(inputDataDate time.Time, targetDataDate *time.Time, granularity string) bool {
+	// If target table has no data, then input data is fresh by default
+	if targetDataDate == nil {
+		return false
+	}
+
+	// We truncate the timestamps to make the comparison at the correct granularity
+	// i.e. input data lagging by two hours is considered stale when granularity is hourly,
+	// but it can still be considered fresh when the granularity is daily.
+	return truncateDate(*targetDataDate, granularity).After(truncateDate(inputDataDate, granularity))
+}
+
 // getRegionForBucket looks up the region name for the given bucket
 func getRegionForBucket(name string) (string, error) {
 	// Any region will work for the region lookup, but the request MUST use
@@ -296,12 +313,7 @@ func main() {
 		}
 
 		// unless --force, don't update unless input data is new
-		// Since targetDataDate comes from the columns, and
-		// inputConf.DataDate comes from the filename, we round to
-		// compare at the time granularity level
-		if *timeGranularity != "stream" &&
-			targetDataDate != nil &&
-			(truncateDate(*targetDataDate, *timeGranularity)).After(truncateDate(parsedInputDate, *timeGranularity)) {
+		if *timeGranularity != "stream" && isInputDataOlder(parsedInputDate, targetDataDate, *timeGranularity) {
 			if *force == false {
 				log.Printf("Recent data already exists in db: %s", *targetDataDate)
 				continue
--- a/main.go
+++ b/main.go
@@ -304,7 +304,7 @@ func main() {
 			(truncateDate(*lastTargetData, *timeGranularity)).After(truncateDate(parsedDate, *timeGranularity)) {
 			if *force == false {
 				log.Printf("Recent data already exists in db: %s", *lastTargetData)
-				return
+				continue
 			}
 			log.Printf("Forcing update of inputTable: %s", inputConf.Table)
 		}
--- a/main.go
+++ b/main.go
@@ -186,7 +186,7 @@ func runCopy(db *redshift.Redshift, inputConf s3filepath.S3File, inputTable reds
 				return err
 			}
 		} else {
-			start, end = startEndFromGranularity(inputConf.DataDate, timeGranularity)
+			start, end = startEndFromGranularity(inputConf.DataDate, timeGranularity, *targetTimezone)
 		}
 		// To prevent duplicates, clear away any existing data within a certain time range as the data date
 		// (that is, sharing the same data date up to a certain time granularity)
@@ -239,7 +239,20 @@ func runCopy(db *redshift.Redshift, inputConf s3filepath.S3File, inputTable reds
 	return nil
 }
 
-func startEndFromGranularity(t time.Time, granularity string) (time.Time, time.Time) {
+func startEndFromGranularity(t time.Time, granularity string, targetTimezone string) (time.Time, time.Time) {
+	// Rotate time if in PT
+	fmt.Println(targetTimezone)
+	if targetTimezone != "UTC" {
+		ptLoc, err := time.LoadLocation(targetTimezone)
+		fatalIfErr(err, "startEndFromGranularity was unable to load timezone")
+
+		_, ptOffsetSec := t.In(ptLoc).Zone()
+		ptOffsetDuration, err := time.ParseDuration(fmt.Sprintf("%vs", ptOffsetSec))
+		fatalIfErr(err, "startEndFromGranularity was unable to parse offset duration")
+
+		t = t.Add(ptOffsetDuration)
+	}
+
 	var duration time.Duration
 	if granularity == "day" {
 		duration = time.Hour * 24
--- a/main.go
+++ b/main.go
@@ -217,7 +217,7 @@ func startEndFromGranularity(t time.Time, granularity string) (time.Time, time.T
 	if granularity == "day" {
 		duration = time.Hour * 24
 	} else {
-		duration = time.Hour * 24 * 7
+		duration = time.Hour
 	}
 
 	start := t.UTC().Truncate(duration)
--- a/main.go
+++ b/main.go
@@ -37,6 +37,7 @@ var (
 	timeGranularity = flag.String("granularity", "day", "how often we expect to append new data")
 	streamStart     = flag.String("streamStart", "", "The start of the streamed data. Only used if granularity='stream'. Used to ensure idempotency")
 	streamEnd       = flag.String("streamEnd", "", "The end of the streamed data. Only used if granularity='stream'. Used to ensure idempotency")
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -37,6 +37,7 @@ var (
 	timeGranularity = flag.String("granularity", "day", "how often we expect to append new data")
 	streamStart     = flag.String("streamStart", "", "The start of the streamed data. Only used if granularity='stream'. Used to ensure idempotency")
 	streamEnd       = flag.String("streamEnd", "", "The end of the streamed data. Only used if granularity='stream'. Used to ensure idempotency")
+	isTargetPST     = flag.Bool("targetPST", false, "Whether or not the target data time column is in PST.")
```
schimmy: PST and PT are not quite the same: https://en.wikipedia.org/wiki/Pacific_Time_Zone
I'd use PT in every instance here, for clarity.

main.go
```diff
@@ -304,7 +304,7 @@ func main() {
 			(truncateDate(*lastTargetData, *timeGranularity)).After(truncateDate(parsedDate, *timeGranularity)) {
 			if *force == false {
 				log.Printf("Recent data already exists in db: %s", *lastTargetData)
-				return
+				continue
```
schimmy: huh - that's a good point...

main.go
```diff
@@ -282,28 +282,28 @@ func main() {
...
+		parsedInputDate, err := time.Parse(time.RFC3339, *dataDate)
 		fatalIfErr(err, fmt.Sprintf("issue parsing date: %s", *dataDate))
-		inputConf, err := s3filepath.CreateS3File(s3filepath.S3PathChecker{}, bucket, *inputSchemaName, t, *configFile, parsedDate)
+		inputConf, err := s3filepath.CreateS3File(s3filepath.S3PathChecker{}, bucket, *inputSchemaName, t, *configFile, parsedInputDate)
 		fatalIfErr(err, "Issue getting data file from s3")
 		inputTable, err := db.GetTableFromConf(*inputConf) // allow passing explicit config later
 		fatalIfErr(err, "Issue getting table from input")
 
 		// figure out what the current state of the table is to determine if the table is already up to date
-		targetTable, lastTargetData, err := db.GetTableMetadata(inputConf.Schema, inputConf.Table, inputTable.Meta.DataDateColumn)
```
schimmy: 👍  on these as well

main.go
```diff
@@ -106,6 +106,23 @@ func truncateDate(date time.Time, granularity string) time.Time {
 	}
 }
 
+// Calculates whether or not input data (s3) is more stale than target data (Redshift)
+// Expects:
+// - inputDataDate corresponds to the s3 data timestamp of the job
+// - targetDataDate is the maximum timestamp of the DB table
+// - granularity indicating how often data snapshots are recorded in the target
+func isInputDataOlder(inputDataDate time.Time, targetDataDate *time.Time, granularity string) bool {
```
schimmy: I would change this to isInputDataStale or similar, based on your other comments - I like that naming convention, I think it's way clearer!

main_test.go
```diff
@@ -18,3 +18,19 @@ func TestTimeGranularity(t *testing.T) {
...
+
+func TestIsInputDataOlder(t *testing.T) {
+	inputDataDate, _ := time.Parse(time.RFC3339, "2017-08-15T14:00:00Z")
+	targetDataDate, _ := time.Parse(time.RFC3339, "2017-08-15T21:00:00Z")
+
+	assert.Equal(t, false, isInputDataOlder(inputDataDate, &targetDataDate, "day", false))
+	assert.Equal(t, true, isInputDataOlder(inputDataDate, &targetDataDate, "hour", false))
+	assert.Equal(t, false, isInputDataOlder(inputDataDate, nil, "hour", false))
+
+	// Simulate Redshift timestamp without time zones that is parsed as UTC but is actually PST
```
schimmy: make sure to also test the day version of this, with two times that would be on different days depending on whether they are PT or UTC!

main.go
evaninja: 👍 0c77f2c

main.go
evaninja: 👍 9f97586

main_test.go
evaninja: Good idea! This seems like the most important functionality for us to test as well - 11365a7

evaninja:
@schimmy really good point that `startEndFromGranularity` changes are required (otherwise the data would have been truncating!)

Updated in 8ea157f - PTAL

main.go
```diff
@@ -37,6 +37,7 @@ var (
 	timeGranularity = flag.String("granularity", "day", "how often we expect to append new data")
 	streamStart     = flag.String("streamStart", "", "The start of the streamed data. Only used if granularity='stream'. Used to ensure idempotency")
 	streamEnd       = flag.String("streamEnd", "", "The end of the streamed data. Only used if granularity='stream'. Used to ensure idempotency")
+	isTargetPT      = flag.Bool("targetPT", false, "Whether or not the target data time column is in PT.")
```
schimmy: I should have thought of this before, but what about if we just pass in the timezone "America/Los_Angeles" as a param, defaulting to "UTC"?

main.go
```diff
@@ -212,12 +243,24 @@ func runCopy(db *redshift.Redshift, inputConf s3filepath.S3File, inputTable reds
...
+		fatalIfErr(err, "startEndFromGranularity was unable to parse offset duration")
+
+		t = t.Add(ptOffsetDuration)
+	}
+
 	var duration time.Duration
 	if granularity == "day" {
 		duration = time.Hour * 24
 	} else {
-		duration = time.Hour * 24 * 7
```
schimmy: why was the default a week previously? Was the default used anywhere? We're changing to hourly as a default here...

main.go
```diff
@@ -106,6 +107,36 @@ func truncateDate(date time.Time, granularity string) time.Time {
...
+		fatalIfErr(err, "isInputDataStale was unable to parse offset duration")
+
+		targetDataDateInUTC := targetDataDate.Add(ptOffsetDuration)
+		targetDataDate = &targetDataDateInUTC
+	}
+
+	// We truncate the timestamps to make the comparison at the correct granularity
+	// i.e. input data lagging by two hours is considered stale when granularity is hourly,
+	// but it can still be considered fresh when the granularity is daily.
+	return truncateDate(*targetDataDate, granularity).After(truncateDate(inputDataDate, granularity))
```
schimmy: nit: I think it is possible to replace truncateDate with startEndFromGranularity. It might be a bit out of scope, but could also simplify understanding of the file.

main.go
schimmy: this also would give us the flexibility of ending on `America/New_York` in the future if we need to / others who might want to use this code

main_test.go
```diff
@@ -10,11 +10,45 @@ import (
...
 	baseTime := time.Date(2017, 7, 11, 12, 0, 0, 0, time.UTC)
 
-	start, end := startEndFromGranularity(baseTime, "day")
+	start, end := startEndFromGranularity(baseTime, "day", false)
 	assert.Equal(t, start, time.Date(2017, 7, 11, 0, 0, 0, 0, time.UTC))
 	assert.Equal(t, end, time.Date(2017, 7, 12, 0, 0, 0, 0, time.UTC))
 
-	start, end = startEndFromGranularity(baseTime, "week")
+	start, end = startEndFromGranularity(baseTime, "hour", false)
+	assert.Equal(t, start, time.Date(2017, 7, 11, 12, 0, 0, 0, time.UTC))
```
schimmy: you should have a minute associated with the `baseTime`, otherwise this isn't testing much!

main_test.go
```diff
@@ -10,11 +10,45 @@ import (
...
+
+	// Simulate timestamps that cross timezones in PT vs UTC
+	baseTime = time.Date(2017, 7, 11, 4, 0, 0, 0, time.UTC)
+
+	start, end = startEndFromGranularity(baseTime, "day", false)
+	assert.Equal(t, start, time.Date(2017, 7, 11, 0, 0, 0, 0, time.UTC))
+	assert.Equal(t, end, time.Date(2017, 7, 12, 0, 0, 0, 0, time.UTC))
+
+	start, end = startEndFromGranularity(baseTime, "day", true)
 	assert.Equal(t, start, time.Date(2017, 7, 10, 0, 0, 0, 0, time.UTC))
```
schimmy: hmm, if we're getting a startEndFromGranularity for a day for PT, shouldn't this give us the PT beginning and ending of the day?
I'd expect `time.Date(2017, 7, 10, 7, 0, 0, 0, time.UTC)` and `time.Date(2017, 7, 11, 16, 0, 0, 0, time.UTC))`, if I'm reading right.

Likely you'll have to do something a bit ugly, see:
- https://stackoverflow.com/questions/36988681/golang-time-time-round-to-day
- https://stackoverflow.com/questions/25254443/return-local-beginning-of-day-time-object-in-go

main.go
evaninja: The default is an hour - having this set to a week was a bug. (Check the README for granularity)

main_test.go
evaninja: Oh wait a minute. I think I remember why I didn't touch `startEndFromGranularity` now - the timestamps coming in are PT, but the output is also in PT. See here https://github.com/Clever/s3-to-redshift/blob/master/main.go#L166 AND https://github.com/Clever/s3-to-redshift/blob/master/redshift/redshift.go#L453-L454:

So it doesn't really matter what timezone we are operating under since the comparisons will be made in the same Redshift column anyways.

Does that sound about right? If so I'll remove these changes to `startEndFromGranularity`

evaninja: tl;dr - the only reason why special hacky logic is necessary in `isInputDataStale` is because that's the ONLY place where PT and UTC timestamps collide. Otherwise, the timezone remains consistent elsewhere.

main_test.go
evaninja: good catch!

evaninja: 0c18619

main_test.go
evaninja: 0c18619

main.go
evaninja: I don't think this makes as much sense anymore - see below comments

main.go
evaninja: sure 117f24f

PR #110, opened 2017-11-10
Increase timeout for student_app_connections

kvigen:
Now that we want to pull user app connections from mongo-to-s3 we have to increase this timeout since it's timing out

**JIRA**:

**Overview**:

**Testing**:

**Roll Out**:


kvigen:
That timeout is just used for the connection (see connect_timeout here https://www.postgresql.org/docs/9.1/static/libpq-connect.html). It's already probably way too large

a-le-jan-dro:
Ohhhhhh so it was effectively useless before 😛  LGTM!

PR #121, opened 2018-04-03
Handle OS Signals Better

Sayan-:
**JIRA**:
https://clever.atlassian.net/browse/IPOC-738

**Overview**:
Similar to: https://github.com/Clever/redshift-vacuum/pull/19.

Let's make sure to clean up in flight queries, etc in case of a clean exit

Note: updating the datadog `sqlmock` makes this entire thing explode 👎. We might want to revisit this in the future if we have bandwidth

**Testing**:
Updated them 👍 

**Roll Out**:
- [x] This repo will auto deploy after you merge. To do a manual deploy, use ark start -e production s3-to-redshift or ark start -e production s3-to-redshift-fast

### DIFF
--- a/redshift/redshift.go
+++ b/redshift/redshift.go
@@ -1,6 +1,7 @@
 package redshift
 
 import (
+	"context"
 	"database/sql"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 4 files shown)


redshift/redshift.go
```diff
@@ -22,16 +23,16 @@ import (
 
 type dbExecCloser interface {
 	Close() error
-	Begin() (*sql.Tx, error)
-	Prepare(query string) (*sql.Stmt, error)
```
Sayan-: Only use the `Prepare` and `Exec` statements from the `tx` not the `db`

bstein-clever:
LGTM!

PR #130, opened 2018-12-04
Print MAU Workflow payloads

a-le-jan-dro:
**Overview**:

Prints out the payloads needed for the next workers (record-keeper and rtp-internal) in the Pulse MAUs workflow being created here: https://github.com/Clever/ark-config/pull/1517

**Testing**:

**Roll Out**:


bstein-clever:
I'm confused by this PR. The payloads are now being hard-coded here in s3-to-redshift?
It feels like that defeats the purpose of having cron jobs for things and having independed job configuration if there're going to be fixed here, and require code changes/deploys to change.

a-le-jan-dro:
Yep! So this is just part of taking small incremental steps and seeing what happens. The idea is for this to not be permanent (unless it works really well somehow 🤔). I'm just trying to get things moving more so the project has more clarity.

a-le-jan-dro:
I copied some code over from `configure` to create a new package. I had originally made changes there but apparently that repo is open source. Anyway, if this looks good and works (I tested manually in dev), we could move it to worker util.

main.go
```diff
@@ -266,19 +266,19 @@ func startEndFromGranularity(t time.Time, granularity string, targetTimezone str
...
-	Force           bool   `json:"force"`
-	DataDate        string `json:"date,required"`
-	ConfigFile      string `json:"config"`
-	GZip            bool   `json:"gzip"`
-	Delimiter       string `json:"delimiter"`
-	TimeGranularity string `json:"granularity,required"`
-	StreamStart     string `json:"streamStart"`
-	StreamEnd       string `json:"streamEnd"`
-	TargetTimezone  string `json:"timezone"`
+	InputSchemaName string `config:"schema"`
```
bstein-clever: This json->config rename is an artifact of the Clever/configure package? I'm not too familiar with it.

bstein-clever:
It looks like there are some merge conflicts that will need to be resolved here.
One question about the latest changes, and I still have my concern from before, but if it's just going be transitional, then I guess it's OK.

a-le-jan-dro:
@bstein-clever why do you still have your concern from before? This gets rid of the payloads being hardcoded in `s3-to-redshift`.

bstein-clever:
Oh, I'd missed that that was changed. In that case, looks good!

PR #132, opened 2019-01-22
Migrate to using workerutil.analyticspipeline for config.

bstein-clever:
**JIRA**:

**Overview**:

**Testing**:

**Roll Out**:

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,6 +2,12 @@
 
 
 [[projects]]
+  branch = "add-analyticspipeline-package"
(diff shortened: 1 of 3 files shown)


Gopkg.lock
```diff
@@ -1,6 +1,12 @@
 # This file is autogenerated, do not edit; changes may be undone by the next 'dep ensure'.
 
 
+[[projects]]
+  branch = "add-analyticspipeline-package"
```
bstein-clever: I don't think we want this branch.

PR #133, opened 2019-01-25
Removed default bucket value and made input required

a-le-jan-dro:
**Overview**:

I don't think this bucket that was set as the default actually exists. If you don't specify a bucket, it throws an error because it is unable to determine the region for it. By making it required, it will instead error because of that.

**Testing**: 💯 

**Roll Out**:


bstein-clever:
Seems reasonable to me.
I agree hard errors are better than throwing a manual error.

PR #135, opened 2019-02-01
S3 data is default read from subfolders.

bstein-clever:
Passion project Wednesday.
Our clever-analytics bucket is a terrible mess.
This puts data into descriptive subfolders for better processing, potentially included with spectrum later on.

I have a script to migrate data, but that comes after the deploy/confirmation this is working.

### DIFF
--- a/s3filepath/s3filepath.go
+++ b/s3filepath/s3filepath.go
@@ -24,12 +24,13 @@ type S3Bucket struct {
 // S3File holds everything needed to run a COPY on the file
 type S3File struct {
 	// info on which file to get
-	Bucket   S3Bucket
-	Schema   string
-	Table    string
-	Suffix   string
-	DataDate time.Time
-	ConfFile string
+	Bucket    S3Bucket
+	Schema    string
+	Table     string
+	Suffix    string
+	DataDate  time.Time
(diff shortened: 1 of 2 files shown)


s3filepath/s3filepath.go
```diff
@@ -54,15 +55,17 @@ func (S3PathChecker) FileExists(path string) bool {
 // GetDataFilename returns the s3 filepath associated with an S3File
 // useful for redshift COPY commands, amongst other things
 func (f *S3File) GetDataFilename() string {
-	return fmt.Sprintf("s3://%s/%s_%s_%s.%s", f.Bucket.Name, f.Schema, f.Table, f.DataDate.Format(time.RFC3339), f.Suffix)
+	return fmt.Sprintf("s3://%s/%s/%s_%s_%s.%s", f.Bucket.Name, f.FilePath, f.Schema, f.Table, f.DataDate.Format(time.RFC3339), f.Suffix)
```
mkmathur: It's not clear what this method does / why it's necessary. Why isn't `f.FilePath` the full path?

bstein-clever: Discussed offline and renamed.

PR #136, opened 2019-02-15
Revert "Remove call to redshift-vacuum"

Sayan-:
Reverts Clever/s3-to-redshift#134

Seeing a lot of:
```
Scanned a large number of deleted rows
```

from `stl_alert_event_log`


bstein-clever:
Sad, but works for me.
Good catch, thanks!

PR #137, opened 2019-02-20
Updated autoscaling to active-percent instead of cpu

a-le-jan-dro:
**JIRA**: No ticket

**Overview**: This worker is always just stuck at 2 because it's not CPU intensive. We should autoscale depending on the business. Right now it's running a pretty much permanent backlog.

**Testing**: 💯 

**Roll Out**:


bstein-clever:
Seems reasonable to me!

PR #141, opened 2019-05-17
Use max query and reduced range scan for more optimized last-value searches.

bstein-clever:
N.B. I changed the order of checks here, and did a few tests locally, so I'm pretty sure about the various error cases, but it's worth double checking those flows with a second pair of eyes.

It turns out that Redshift doesn't have good optimizations for queries of the form:
`select max(time) from <table>` when `time` is the sort key for the table, as the min/max-chunks on blocks are only used for range reductions, and this query is otherwise a full-table scan. :(
By adding a where clause like so:
`select max(time) from <table> where time > <1 day ago>`
we can get significant performance because we're looking over a smaller range.

Because it's not guaranteed that we always have results in a specific time range (failed jobs, infrequent firehose logs, etc.) we do multiple passes on increasingly larger time ranges until we get a hit.
In worst-case, we default to the entire table scan again, which is no worse than before.

I did confirm that querying for values that are out of range is extremely fast:
```
prod=#  select time from historical.managed_login_facts_vw_stream where time > '2020-05-01' order by time desc limit 1;
 time 
------
(0 rows)

Time: 30.998 ms
prod=#  select time from historical.managed_login_facts_vw_stream where time > '2020-05-01' order by time desc limit 1;
 time 
------
(0 rows)

Time: 386.647 ms
```
So even in this worst-case, we shouldn't be affecting overall runtime significantly.

### DIFF
--- a/redshift/redshift.go
+++ b/redshift/redshift.go
@@ -16,7 +16,7 @@ import (
 	multierror "github.com/hashicorp/go-multierror"
 	// Use our own version of the postgres library so we get keep-alive support.
 	// See https://github.com/Clever/pq/pull/1
-	_ "github.com/Clever/pq"
+	"github.com/Clever/pq"
 
 	"github.com/Clever/s3-to-redshift/s3filepath"
 )
@@ -60,6 +60,20 @@ type ColInfo struct {
 	PrimaryKey  bool   `yaml:"primarykey"`
 	DistKey     bool   `yaml:"distkey"`
 	SortOrdinal int    `yaml:"sortord"`
+}
+
+type rangeQuery int
+
+const (
+	rangeAll rangeQuery = iota
+	rangeYear
+	rangeMonth
+	rangeWeek
+	rangeDay
+)
+
+func rangeQueryString(r rangeQuery) string {
+	return []string{"", "YEAR", "MONTH", "WEEK", "DAY"}[r]
 }
 
 const (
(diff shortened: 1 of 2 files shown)


redshift/redshift.go
```diff
@@ -62,6 +62,20 @@ type ColInfo struct {
...
+const (
+	rangeNone rangeQuery = iota
+	rangeYear
+	rangeMonth
+	rangeWeek
+	rangeDay
+)
+
+func rangeQueryString(r rangeQuery) string {
+	return []string{"", "YEAR", "MONTH", "WEEK", "DAY"}[r]
```
mkmathur: nice!

redshift/redshift.go
```diff
@@ -62,6 +62,20 @@ type ColInfo struct {
 	SortOrdinal int    `yaml:"sortord"`
 }
 
+type rangeQuery int
+
+const (
+	rangeNone rangeQuery = iota
```
mkmathur: `rangeNone` means that we scan the entire table, right? I would maybe call that `rangeAll`.

redshift/redshift.go
```diff
@@ -212,18 +226,46 @@ func (r *Redshift) GetTableMetadata(schema, tableName, dataDateCol string) (*Tab
...
 
+// MaxTime returns the maximum value for the time field in the specified table
+func (r *Redshift) MaxTime(fullName, dataDateCol string) (time.Time, error) {
+	return r.maxTime(fullName, dataDateCol, rangeDay)
+}
+
+// maxTime is a helper function to scan progressively larger ranges of time to get more optimized
+// max(time) queries - redshift doesn't have good optimizations for max on sort-keyed columns.
+func (r *Redshift) maxTime(fullName, dataDateCol string, rangeLimit rangeQuery) (time.Time, error) {
+	lastDataQuery := fmt.Sprintf(`SELECT "%s" FROM %s ORDER BY "%s" DESC LIMIT 1`,
```
mkmathur: should be max

bstein-clever:
Good catch, thanks! Fixed.

redshift/redshift.go
bstein-clever: I like that. Updated.

redshift/redshift.go
bstein-clever: Good catch, thanks! Fixed.

PR #145, opened 2019-06-11
All jobs should kick off a vacuum - e.g. streams, at minimum a delete.

bstein-clever:
I recommend turning off whitespace diffs on this one.


bstein-clever:
cc @Sayan-, related to our discussion about vacuums on loads.

PR #146, opened 2019-07-19
remove compupdate flag

Sayan-:
since we aren't creating new tables the compupdate flag doesn't give us any benefits

### DIFF
--- a/redshift/redshift.go
+++ b/redshift/redshift.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/Clever/pathio"
 	multierror "github.com/hashicorp/go-multierror"
+
 	// Use our own version of the postgres library so we get keep-alive support.
 	// See https://github.com/Clever/pq/pull/1
 	"github.com/Clever/pq"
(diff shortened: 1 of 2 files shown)


bstein-clever:
LGTM, thanks for the fast turnaround!

bstein-clever:
Actually, can we make this `PRESET` instead of removing it? https://docs.aws.amazon.com/redshift/latest/dg/copy-parameters-data-load.html

Sayan-:
I don't think we want the `PRESET`

>When COMPUPDATE is PRESET, the COPY command chooses the compression encoding for each column if the target table is empty; even if the columns already have encodings other than RAW. Currently specified column encodings can be replaced. Encoding for each column is based on the column data type. No data is sampled. Amazon Redshift automatically assigns compression encoding as follows:

since we `DELETE` -> `COPY` we would be redoing compression every job

bstein-clever:
Oh, interesting find. I think there are two thoughts I'm trying to balance here:
1) What if the data does meaningfully change? The fact that the column encodings can be redone if the table is empty is kind of cool.
2) This does (occasionally) come up when a new table is created. And it would be neat if we got the correct compression in that situation, even if we skip it on reload.

Thoughts?

Sayan-:
IMO the way we're currently leveraging this worker `DELETE`->`COPY` >>>>>> `CREATE` -> `COPY`. We're optimizing for a case we don't really need to be caring about.

I also think if our data is meaningfully changing we probably have bigger problems...

bstein-clever:
Well, we do often add columns (not sure how that affects this), and we do *add* new content (e.g. glue, historical tables) regularly.

Sayan-:
Taking a quick step back, we saw some pretty _major_ compression changes when we ran the column encoding script. Considering that happened on tables with the `COMPUPDATE` flag enabled, I think it probably makes more sense to make it a scheduled task every quarter / half year since the automated process isn't meeting our requirements

bstein-clever:
Chatted offline about this. I think re-compressing all our tables is too big of a task to do every quarter/half, and I think there wasn't significant change in most of the (already) compressed tables - it was a difference in that a number of them had no compression set.

My question, in particular, had to do with adding new columns, but I suspect that's not worth over-rotating on right now. I'd love to get this out soon - let me know if you need me to take it over, or if you think you can do these (and the relevant redbox updates) before you leave for vacation.

Sayan-:
Sounds good - happy to let you take over from here!

PR #155, opened 2020-04-17
Add thresholds for analyze

jansenclever:
**JIRA**:
https://clever.atlassian.net/browse/IPOC-1345

**Overview**:
`redshift-cleanup` already defaults to vacuum full (so we were running it all the time regardless, but this will make it more explicit. hmlfvs analyze was taking up to a hour at some points and we run it every hour, so it clogs up the pipeline.

Also, truncate would naturally hit a 100% threshold anyway so this really just makes it so we skip extra analyzes that might not be necessary. (we analyze schemas at 0.001% every day already).

Related to https://github.com/Clever/redshift-cleanup/pull/44

Alternatively, we can make the non-truncate skip analyzing

### DIFF
--- a/main.go
+++ b/main.go
@@ -219,10 +219,10 @@ func runCopy(
 		cleanupArgs := map[string]string{
 			"targets":     inputConf.Schema + `."` + inputTable.Name + `"`,
 			"vacuum_mode": "delete",
-		}
-		// If we truncated, run an analyze as well
-		if truncate {
-			cleanupArgs["analyze_mode"] = "full"
+			// If we truncated, analyze will run regardless since 100% of the rows have changed. Otherwise,
+			// only analyze if we've changed enough rows (threshold > 1%)
+			"analyze_mode":      "full",
+			"analyze_threshold": "1",
 		}
 
 		payload, err := json.Marshal(cleanupArgs)


main.go
```diff
@@ -220,9 +220,13 @@ func runCopy(
...
 		}
-		// If we truncated, run an analyze as well
+		// If we truncated, run an analyze as well (threshold 0.001%). Otherwise,
+		// only analyze if we've changed enough rows (> 10%, redshift default)
+		cleanupArgs["analyze_mode"] = "full"
 		if truncate {
-			cleanupArgs["analyze_mode"] = "full"
+			cleanupArgs["analyze_threshold"] = "0.001"
+		} else {
+			cleanupArgs["analyze_threshold"] = "10"
```
jansenclever, line 229: The 10% default got us into trouble in the past with hmlfvs because it was longer than 28 days of data.  We could make this smaller but we're also running it at 0.001% every night so maybe this isn't really necessary.

jansenclever: Or we can use a smaller threshold like 1%, that won't be triggered on every run but would still run if our nightly analyze fails for some reason.

bstein-clever: I think we've got a lot of tables that could benefit from a smaller percentage. I like the idea of having a default that's a bit more optimized than 10%. I made it so small since I wanted to make sure we weren't at risk of running into the 30-day window for hmlfvs again any time soon.

bstein-clever: That said, I think this distinction might not be necessary. It we truncate the table, then we've changed 50% of the data, so any (reasonable) analyze should do it. It was just that it's important to reprocess, unlike historical tables, where there are probably fewer underlying changes.

bstein-clever:
This will have the side-effect of running more analyzes than previously - I believe the previously missing else block was intentional.

I'm not sure if that's what we want to be doing here.

jansenclever:
Hmm, I think redshift-analyze defaulted to "full" if it wasn't provided, so we were always analyzing.
https://github.com/Clever/redshift-cleanup/blob/master/main.go#L61

bstein-clever:
Yeah, fair point. I think this solves what I was probably trying to do at the time in a better way.
And since I forgot to say it: LGTM.

main.go
jansenclever: Agreed, I was thinking that too when adding this but I wanted it to be more explicitly written out. I will put it directly into the blob with a more detailed comment, should achieve the same thing.

jansenclever: I'll use 1% as the threshold

PR #160, opened 2020-08-25
Write latency data to data and firehose for easy lookup.

bstein-clever:
**JIRA**:

**Overview**:

**Testing**:

**Roll Out**:


redshift/redshift.go
```diff
@@ -479,6 +482,48 @@ func (r *Redshift) Copy(tx *sql.Tx, f s3filepath.S3File, delimiter string, creds
...
+
+	// Insert a row for the latencies table if it doesn't already exist.
+	// We do this outside of the transaction, so that we can get a row-level lock with the next command.
+	_, err := r.ExecContext(r.ctx, fmt.Sprintf(
+		`INSERT INTO latencies (name) (
+				SELECT name FROM latencies
+			UNION
+				SELECT '%s' AS name
+			EXCEPT
+				SELECT name FROM latemcoes
```
jansenclever, line 498: Typo? `FROM latencies`

redshift/redshift.go
```diff
@@ -479,6 +482,48 @@ func (r *Redshift) Copy(tx *sql.Tx, f s3filepath.S3File, delimiter string, creds
...
+	_, err = tx.ExecContext(r.ctx, fmt.Sprintf(
+		"UPDATE latencies SET last_update = current_timestamp WHERE name = '%s'",
+		dest))
+	if err != nil {
+		return err
+	}
+
+	logger.GetLogger().InfoD("analytics-run-latency", kvlogger.M{
+		"destination":  dest,
+		"previous_run": t.Time,
```
jansenclever, line 521: Why do we want the previous time? Won't that be in the event stream anyway?

jansenclever: Oh, I guess this gives us the actual current_timestamp rather than the log time.

jansenclever: We could select it out of the transaction too if we want?

jansenclever: Oh, but we don't know if this ends up succeeding until later, haha

PR #168, opened 2020-09-08
Create new s3-to-redshift-key-metrics resource

a-le-jan-dro:
**Overview**: Hoping to have some workflows avoid the long s3-to-redshift queue, much like we do with some rtp steps using redshift-to-postgres-key-metrics.

Goes with https://github.com/Clever/ark-config/pull/2742


jansenclever (approved):
lgtm

PR #169, opened 2020-10-01
Add support for skipping loads.

bstein-clever:
This should really only be used in conjunction with job debouncing.

**JIRA**:
https://clever.atlassian.net/browse/DEIP-171

**Overview**:
Hot on the tail of https://github.com/Clever/connections-to-s3/pull/104

Now that it works, we can add debouncing to other services.
s3-to-redshift needs to have an arg from the debouncing, so it knows when a load would be invalid,
so this has to merge before the mongo-to-s3 or record-keeper jobs.

**Testing**:

**Roll Out**:

### DIFF
--- a/main.go
+++ b/main.go
@@ -290,6 +290,7 @@ type payload struct {
 	StreamStart     string `config:"streamStart"`
 	StreamEnd       string `config:"streamEnd"`
 	TargetTimezone  string `config:"timezone"`
+	SkipLoad        bool   `config:"skipLoad"`
 }
 
 // This worker finds the latest file in s3 and uploads it to redshift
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -290,6 +290,7 @@ type payload struct {
 	StreamStart     string `config:"streamStart"`
 	StreamEnd       string `config:"streamEnd"`
 	TargetTimezone  string `config:"timezone"`
+	SkipLoad        bool   `config:"skip-load"`
```
jansenclever, line 293: Nit: Should we make this skipLoad to match the case of streamStart/streamEnd? Though I don't what our standard is, seems like it's all over the place

jansenclever (approved):
LGTM

main.go
jansenclever: (I don't feel strongly about this)

jansenclever: I guess `skip-debounce` also has a dash but we didn't have two word config params in the other repros so I didn't think about it then.

bstein-clever: Chatted offline. Switched it to be camelCase, and consistent across the repos with skipDebounce.

PR #180, opened 2022-05-16
update pod_config.group to reduce cross region dependencies

tnsardesai:
Same as https://github.com/Clever/matchmaker-spawner/pull/39


jansenclever:
This reads from a us-west-1 s3 bucket and puts it into a us-west-2 redshift cluster

tnsardesai:
Fine to close this right now. We might revisit this later if necessary

